Issue #6, opened 2018-05-16
Possible bug in meshopt_remapVertexBuffer()


corporateshark:
`meshopt_remapVertexBuffer()` works only for unindexed vertices. Is it by design?

zeux:
It should work for indexed vertices as well. Can you clarify why you think it doesn't?

zeux:
meshopt_remapVertexBuffer should be called on a remap table that is generated by meshopt_generateVertexRemap; the table maps old vertex indices to new vertex indices, e.g. if you started from a mesh with 4 vertices 2 of which are redundant, remap[i] will be defined for i=0,1,2,3, and will take values 0,1,2.

vertex_count parameter in meshopt_remapVertexBuffer is the *old* vertex count. So if you're remapping an indexed mesh above, it'd be 4, not 3. The function will then move 3 unique vertices to wherever they need to be. Does that make sense?

(I must admit this specific part was confusing to at least one other person so this needs some documentation improvements...)

PR #7, opened 2018-05-18
Fixed documentation issue which caused API ambiguity

corporateshark:
See #6 for details.


zeux:
Thanks! I was also thinking about changing the interface to pass *both* vertex counts - that way the assertion and the loop would use different values - but it's not ideal since technically one of the arguments would be redundant. This is great for now.

PR #17, opened 2019-01-12
Optimize meshopt_computeMeshletBounds

gwihlidal:
Optimize meshopt_computeMeshletBounds by passing in pointer to meshlet instead of by value (636 bytes)

### DIFF
--- a/demo/main.cpp
+++ b/demo/main.cpp
@@ -679,7 +679,7 @@ void meshlets(const Mesh& mesh)
 	double startc = timestamp();
 	for (size_t i = 0; i < meshlets.size(); ++i)
 	{
-		meshopt_Bounds bounds = meshopt_computeMeshletBounds(meshlets[i], &mesh.vertices[0].px, mesh.vertices.size(), sizeof(Vertex));
+		meshopt_Bounds bounds = meshopt_computeMeshletBounds(&meshlets[i], &mesh.vertices[0].px, mesh.vertices.size(), sizeof(Vertex));
 
 		// trivial accept: we can't ever backface cull this meshlet
 		accepted += (bounds.cone_cutoff >= 1);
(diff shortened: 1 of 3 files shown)


zeux:
This effectively reverts https://github.com/zeux/meshoptimizer/commit/925ff15122ce4b77ce2ed1951d1337ef059814b5#diff-82d115d4e3c804084031200d2dc7dc62; the premise of that change was that the callers in C/C++ would prefer pass-by-value since it's more intuitive than passing a singular pointer (in C++ I'd pass a const ref, but the interface is C89), and the performance penalty is negligible.

My mild preference is to keep the pass by value, but I'll need to double check what the performance impact is (if it's noticeable then we should definitely pass by pointer), whether any languages have issues with FFI when passing structs by value (probably not?) and also I'm not sure how idiomatic it is to pass by value vs by singleton pointer in C in general...

zeux:
Here's data from sanMiguel.obj on a revised version of this change where meshopt_computeMeshletBoundsPtr calls computeMeshletBounds (and does the copying-to-stack required):

image

So this suggests that out of 242ms generating bounds data for meshlets, we spend 3ms copying, making the copy overhead ~1%. That's with the default meshlet parameters (64 max vertices, 126 max triangles), however my sanMiguel model has very poor vertex-triangle ratio, so the average meshlet ends up having 26 triangles which basically maximizes the copy cost relative to copying cost - on more reasonable models the overhead should be below 1%.

So unless there's any issues with FFI (I've double checked and it seems like well known FFIs all support passing structs by value?), I'd prefer to close this - let me know if my reasoning feels off!

jamesdolan:
No issue, just popped up in static analysis.

zeux:
Since this came up again in a linked issue I'm having second thoughts and leaning to merging this. The performance impact is minimal but it seems like it's better to pass meshlet by pointer for now. Eventually I kinda want to redesign the meshlet API to not have a structure with hardcoded limits but this complicates the meshlet builder - for now pointer is fine I guess. Let me try to see if I can reopen & merge this...

PR #50, opened 2019-07-23
gltfpack: Add options to keep mesh identity and node hierarchy

### LINKED ISSUE #48: gltfpack does not preserve node hierarchy
Is there some option to keep the "body" nodes in the image below:
image

because running gltfpack without parameters results in a flat node list:

image

this makes it impossible to move the bodies

zeux:
This change introduces two new gltfpack options:

- `-km` disables mesh merging and keeps meshes attached to their original nodes. This makes sure that each mesh can be individually transformed and modified by the application
- `-kn` disables node hierarchy simplification. This makes sure that the application can reference nodes that were originally present in the file, for example to attach external objects

In addition, node names are no longer being removed since they don't result in a noticeable file size increase and it's easier to reason about nodes as just being kept or not - if the node is preserved without modifications, it keeps the name as well.

Fixes #48.

### DIFF
--- a/tools/gltfpack.cpp
+++ b/tools/gltfpack.cpp
@@ -75,6 +75,7 @@ struct Settings
 	int anim_freq;
 	bool anim_const;
 
+	bool keep_nodes;
 	bool keep_meshes;
 
 	bool compress;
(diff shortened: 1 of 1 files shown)


zeux:
Note: after discussing the usecases a bit more this PR ended up being a bit too aggressive so it won't be merged - it's superseeded by PR #51 (the issue is that introduced options are suboptimal for the applications).

PR #52, opened 2019-08-05
Support BUILD_SHARED_LIBS

l0calh05t:
Adds the ability to build as a shared library using the global CMake option BUILD_SHARED_LIBS


zeux:
Thanks!

PR #54, opened 2019-08-05
Add a CMake install target

l0calh05t:
Adds an install target with config-/export-based `find_package` support. Usage example for the installed version:

```cmake
find_package(meshoptimizer REQUIRED CONFIG)
target_link_libraries(<some_target> PRIVATE meshoptimizer::meshoptimizer)
```

Note that the current version must be passed to `project` (also, `cmake_minimum_required` should be called *before* `project`).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,5 @@
-project(meshoptimizer)
-cmake_minimum_required(VERSION 3.0)
+cmake_minimum_required(VERSION 3.0 FATAL_ERROR)
+project(meshoptimizer VERSION 0.12 LANGUAGES CXX)
 
 option(BUILD_DEMO "Build demo" OFF)
 option(BUILD_TOOLS "Build tools" OFF)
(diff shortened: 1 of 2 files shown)


zeux:
If `BUILD_TOOLS` is set, does the install target install `gltfpack` to the target binary folder? Also, does this setup work for doing a global install on Linux-like systems?

l0calh05t:
No, `gltfpack` is not currently installed. But this would be easy to change. Should the same be done for `demo` as well?

It should work fine for a global install on POSIX systems. The default `CMAKE_INSTALL_PREFIX` is `/usr/local` on POSIX, so libraries would be installed to `/usr/local/lib`, includes to `/usr/local/include`, etc.

zeux:
Probably not for demo; gltfpack is meant to be a user facing tool whereas demo just contains tests and sample code.

l0calh05t:
Done. `gltfpack` is now installed in bin. For Linux it might make sense to set an `$ORIGIN`-relative `RPATH`.

zeux:
Thanks! I'll play with this locally on Linux, we can add RPATH later if we need to.

l0calh05t:
You're welcome. I did try it out a bit on WSL, and I do think an `RPATH` of `$ORIGIN/../lib` should be set so that `gltfpack` will run no matter what the prefix path is without having to modify `LD_LIBRARY_PATH`. It's a single-line change, so it would be easy to add.

PR #65, opened 2019-10-02
gltfpack: Add simplification options

### LINKED ISSUE #60: Add simplify option to gltfpack
Thank you for this super useful library! I would like to further optimize models by running simplification. It would be really helpful if `gltfpack` allowed me to do this one shot.

For example, `-si 0.4` could reduce the poly count to 40% of the original.

` ./gltfpack -i glTF-Sample-Models/2.0/BarramundiFish/glTF/BarramundiFish.gltf  -o ./glTF/BarramundiFish-simple.gltf -si 0.4`

zeux:
This change adds two new command-line arguments, -si R (to specify the simplification ratio) and -sa (to enable aggressive simplification). Each mesh in the output is simplified to the given ratio; when aggressive simplification is enabled, the target ratio is reached disregarding change in appearance.

Supporting this for point cloud meshes required a new algorithm, `meshopt_simplifyPoints`; this uses a variant of sloppy simplifier with the quadric error based on distances to points. This is an experimental addition and will probably need to be refined (in particular, attribute error is pretty important for point clouds).

In the future the aggressive mode can be refined a bit further by trying to simplify the mesh with the default simplifier on a subset of attributes in some cases (e.g. trying to preserve UV seams but ignoring normal/tangent seams, which can maintain appearance a bit better compared to the sloppy simplifier).

Fixes #60.

### DIFF
--- a/README.md
+++ b/README.md
@@ -288,7 +288,7 @@ You can then run the resulting command-line binary like this (run it without arg
 gltfpack -i scene.gltf -o scene.glb
 ```
 
-gltfpack substantially changes the glTF data by optimizing the meshes for vertex fetch and transform cache, quantizing the geometry to reduce the memory consumption and size, merging meshes to reduce the draw call count, quantizing and resampling animations to reduce animation size and simplify playback, and pruning the node tree by removing or collapsing redundant nodes.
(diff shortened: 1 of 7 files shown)


zeux:
Source:
image
image

After `./gltfpack -si 0.4`:

image
image

PR #68, opened 2019-10-09
gltfpack: Rework compression extension to allow for uncompressed fallback

zeux:
This change reworks the extension definition so that instead of using bufferOffset/bufferLength/buffer from the bufferView, the extension JSON specifies these directly.

This is important for two reasons:

1. This allows us to produce conformant compressed files - previously, bufferLength in the buffer view violated the spec (since the accessor needed to read more bytes from the buffer view), and it wasn't clear how to resolve this

2. This allows us to produce compressed files with uncompressed fallbacks (`-cf` option). In this mode, the `.bin`/`.glb` data is still using optimally packed compressed data, but we generate an extra `.fallback.bin` file with uncompressed data. Loaders that support compression don't need to load this buffer, but the files are compatible with loaders that don't support compression extension.

When generating compressed files without fallback (`-c`), we generate a "dummy" buffer with no URI. This is necessary because, since bufferView.buffer is a required schema element, this seems to be the only reasonable option of generating files that conform to glTF specification.

This is a breaking change in the compression extension specification so this change updates loaders and has to update the demo `.glb` file.


zeux:
There are some questions regarding the exact approach taken here with fallback buffers in absence of the said fallback - it's not clear if dummy buffer references are optimal. There's a tracking issue for this in the spec repository, https://github.com/KhronosGroup/glTF/issues/1684 - depending on the way it is resolved, this PR may or may not need changes. Since this is a large change to the JSON schema, I would rather do it once so I'm going to wait to merge this for a bit.

PR #73, opened 2019-11-03
Enable WASM SIMD support

zeux:
This PR completes a few remaining items to be able to enable WASM SIMD:

- We now build two .wasm blobs, with and without SIMD support
- At runtime, we select the SIMD blob if a small detector module (that uses SIMD and bulk-memory instructions) validates
- Instead of manually initializing global table, we use `_start` symbol that does it for us

The .js blob gets noticeably larger with this change, but the increase post-deflate is reasonable:

.js: 8406 -> 18292 (2.17x)
.js.gz: 3879 -> 6944 (1.79x)

Raw .wasm files stats are as follows:

decoder_base.wasm 4885
decoder_base.wasm.gz 2117
decoder_simd.wasm 7103
decoder_simd.wasm.gz 2135

The more modest increase in .js.gz size is probably due to index codec implementation being shared between the two modules. We also use bulk-memory for the second module which cuts the .wasm size a bit (by ~900 bytes).

As per #72, on Chrome Canary SIMD support results in ~3x speedup when decoding vertex data so 2 KB of download size seems a reasonable price to pay.

Note that this PR requires Emscripten 1.39.2 (unreleased) for LLVM fixes; it will be merged one this version of Emscripten is actually live.


zeux:
Ok the extra size growth was bothering me so I decided to explore alternative storage strategies.

As you can see in the PR message, .wasm files compress *really* well but we clearly lose a lot of size post-deflate because of Base64 conversion. This makes sense - Base64 switches byte alignment around so if before Base64 your binary had recognizable byte pattern matches, after Base64 it will only have recognizable byte pattern matches if in the source binary the matches were aligned on 3b boundary.

Since we'd like to keep just one file, we can resolve this by storing hex string data instead of Base64. It increases the size of the raw .js blob but decreases the size of post-deflate blob. This is because byte matches in source file are replaced with 2-byte matches in destination file, so they are still picked up by gzip.

Sizes are as follows (scalar: before this PR; dual: two blobs after this PR):

Base64, scalar: .js 8799, .js.gz 4049
Base64, dual: .js 18292, .js.gz 6943
Hex, scalar: .js 12132, .js.gz 3472
Hex, dual: .js 26359, .js.gz 4723

So it looks like it's not only the case that hexadecimal notation compresses better, it's also easier for gzip to reuse data between simd and baseline versions so the total relative size impact is smaller.

abrown:
@zeux, how were you building this? I'm trying `make js/meshopt_decoder.js` with emscripten 1.39.2 but am getting errors like:
```
emcc src/vertexcodec.cpp src/indexcodec.cpp -O3 -DNDEBUG -s EXPORTED_FUNCTIONS='["_meshopt_decodeVertexBuffer","_meshopt_decodeIndexBuffer","_sbrk"]' -s ALLOW_MEMORY_GROWTH=1 -s TOTAL_STACK=24576 -s TOTAL_MEMORY=65536 -o build/decoder_base.wasm
emcc src/vertexcodec.cpp src/indexcodec.cpp -O3 -DNDEBUG -s EXPORTED_FUNCTIONS='["_meshopt_decodeVertexBuffer","_meshopt_decodeIndexBuffer","_sbrk"]' -s ALLOW_MEMORY_GROWTH=1 -s TOTAL_STACK=24576 -s TOTAL_MEMORY=65536 -o build/decoder_simd.wasm -munimplemented-simd128 -mbulk-memory
clang-6.0: error: unknown argument: '-munimplemented-simd128'
clang-6.0: error: unknown argument: '-mbulk-memory'
```
Do I need a different version of LLVM?

zeux:
@abrown 1.39.2 should work out of the box. I'm surprised you get an error from clang-6.0. Any chance your ~/.emscripten file has LLVM_ROOT that points into a custom LLVM build?

zeux:
The way I built this yesterday after 1.39.2 release was to install 1.39.2 through emsdk:

```
emsdk install latest
emsdk activate latest
source emsdk_env.sh
```

And then call make like you did:

```
make -j2 js/meshopt_decoder.js
```

abrown:
That worked for me when I ran it natively; there must be something wrong with the Docker image I was using.

PR #78, opened 2019-11-18
gltfpack: Impement support for Basis texture conversion

zeux:
When -tb is specified, we run all textures from the input asset through
Basis, and place the results into .basis folder next to the scene asset.

Embedding textures is supported using image/basis MIME type although
this is not supported by any glTF renderers - the plan is to migrate this code
to use KTX2 once there's some support for it anywhere.

Also normal encoding quality is pretty bad - it requires splitting the image into
two channels, which again isn't supported by any existing renderers.

### DIFF
--- a/tools/gltfpack.cpp
+++ b/tools/gltfpack.cpp
@@ -2706,15 +2706,22 @@ bool readFile(const char* path, std::string& data)
 	}
 
 	data.resize(length);
-
-	if (fread(&data[0], 1, data.size(), file) != data.size())
-	{
-		fclose(file);
-		return false;
-	}
-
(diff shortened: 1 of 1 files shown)


zeux:
Note: this change *only* supports .png inputs (because of basisu) and *only* supports .basis outputs (because of limitations of KTX2 support in the ecosystem atm). I'm planning to merge this as is but note that .basis support will likely *not* stay forever - since glTF is standardizing on .ktx2 container format, this is just used to test out the image conversion/embedding code. I'll probably submit temporary code to demo/GLTFLoader.js to support this workflow in the meantime though.

PR #87, opened 2019-12-15
Fix build on Android (CMake < 3.7).

Themaister:
VERSION_GREATER_EQUAL is not available on older CMake versions.


Themaister:
Reference: https://github.com/glfw/glfw/issues/1584

zeux:
Thanks!

PR #89, opened 2019-12-17
tools/lodviewer: Add `lodviewer` to CMakeLists.txt

corporateshark:


CMakeLists.txt
```diff
@@ -66,6 +66,16 @@ if(BUILD_TOOLS)
         string(CONCAT RPATH "$ORIGIN/../" ${CMAKE_INSTALL_LIBDIR})
         set_target_properties(gltfpack PROPERTIES INSTALL_RPATH ${RPATH})
     endif()
+
+    if(EXISTS glfw)
```
zeux: Is that the right way to discover glfw or is there some way to use find_package to do this? Not sure if this just checks for the folder.

corporateshark: If we are building GLFW from the specific sources (which is the case for Windows), then this is pretty much in-line with the official documentation: https://www.glfw.org/docs/latest/build_guide.html (section `With CMake and GLFW source`).

The `find_package` approach is error-prone and unstable on Windows.

zeux: I would need to look into this later; I'm not comfortable merging this as is because it feels very single-use and against the best practices of CMake that I know. The manual you link describes the find_package route as well, so I would ask if find_package can work on Windows given some external configuration.

If we are going down the source path route I'd want the path to be explicitly configurable and unset by default so that there's no assumption that glfw is in meshoptimizer's source tree.

corporateshark: What if we check the user provided path first, then (if user provided no custom path) we do `find_package()`, and then we check the hardcoded `glfw` folder, and then, if everything else has failed, just skip building `lodviewer`? IMHO, this should cover Linux/Windows hosts pretty well.

zeux: This seems reasonable. We could also start by simply changing the hardcoded glfw path to a (required) user input, and do the find_package dance separately.

I'm kinda hoping that long term CMake will evolve to make find_package route the only one necessary; it looks like today this might be feasible with vcpkg but it's not integrated into VS nicely so may not be super compelling just yet.

corporateshark:
Added trivial (canonical) `find_package()` and the string option `BUILD_TOOLS_GLFW_FOLDER_NAME`.

Usage example:
```
cmake .. -G "Visual Studio 16 2019" -A x64 -DBUILD_DEMO=ON -DBUILD_TOOLS=ON -DBUILD_TOOLS_GLFW_FOLDER_NAME=glfw3
```

CMakeLists.txt
```diff
@@ -67,14 +68,23 @@ if(BUILD_TOOLS)
...
     endif()
 
-    if(EXISTS glfw)
+    find_package(glfw3 3.3 QUIET)
+
+    if (glfw3_FOUND)
+        message(STATUS "Found GLFW3 via find_package()")
+        add_executable(lodviewer tools/lodviewer.cpp tools/meshloader.cpp)
+        target_link_libraries(lodviewer ${GLFW3_LIBRARY} meshoptimizer)
+    elseif(EXISTS ${BUILD_TOOLS_GLFW_FOLDER_NAME})
```
zeux: This should probably be renamed to smth like MESHOPT_GLFW_PATH

zeux: Also, would it be possible to use a defined check vs exists? I'd like it to be the case that if the option is specified, we *always* act on it - letting CMake produce errors if the path is incorrect.

CMakeLists.txt
```diff
@@ -67,14 +68,23 @@ if(BUILD_TOOLS)
...
         set(GLFW_BUILD_EXAMPLES OFF CACHE BOOL "")
         set(GLFW_BUILD_TESTS OFF CACHE BOOL "")
         set(GLFW_BUILD_DOCS OFF CACHE BOOL "")
         set(GLFW_INSTALL OFF CACHE BOOL "")
-        add_subdirectory(glfw)
+        add_subdirectory(${BUILD_TOOLS_GLFW_FOLDER_NAME})
         add_executable(lodviewer tools/lodviewer.cpp tools/meshloader.cpp)
         target_link_libraries(lodviewer glfw meshoptimizer)
+    else()
+        message(STATUS "GLFW3 not found. LODViewer disabled.")
```
zeux: I don't think we need this diagnostics, or the one above when the path is specified explicitly. It probably still makes sense to emit one for find_package, I don't remember if it emits it by itself.

CMakeLists.txt
```diff
@@ -67,14 +68,23 @@ if(BUILD_TOOLS)
         set_target_properties(gltfpack PROPERTIES INSTALL_RPATH ${RPATH})
     endif()
 
-    if(EXISTS glfw)
+    find_package(glfw3 3.3 QUIET)
+
+    if (glfw3_FOUND)
+        message(STATUS "Found GLFW3 via find_package()")
+        add_executable(lodviewer tools/lodviewer.cpp tools/meshloader.cpp)
+        target_link_libraries(lodviewer ${GLFW3_LIBRARY} meshoptimizer)
```
zeux: If the code below declared GLFW3_LIBRARY variable locally, these two lines (that will change whenever the set of source files required changes) could be moved to be outside of GLFW discovery code.

zeux:
Thanks!

PR #97, opened 2020-01-27
Index codec v2

zeux:
This PR introduces a set of changes required to get better compression for large meshes.

First, in addition to the cache optimization that minimizes ACMR we now can perform an optimization that favors strip-like sequences via meshopt_optimizeVertexCacheStrip. This can result in substantially shorter strips so it's a great idea for stripification (e.g. on boi.obj with restarts, the index buffer gets reduces by 1.3x with this), and it also improves index and vertex compression.

Naturally this negatively impacts ACMR, typically resulting in ~10% worse vertex reuse rates compared to meshopt_optimizeVertexCache.

Second, in order to use this order most effectively, index codec needs to support strip-like sequences. When a strip goes through the previously unexplored parts of the mesh, our existing encoding works well - however, when it returns back, every other triangle ends up reusing a previously met vertex with sequentially decreasing indices (since they sequentially increased in the forward run). For this to work better, index codec now supports special encoding for "last" vertex in addition to "next".

Third, since we need to change the index encoding format anyhow, this change introduces support for index restarts that reset "next" back to 0. This is really important to be able to merge multiple index sequences into one without sacrificing compression ratio, which is useful for gltfpack.

To use the new index format, the application needs to call `meshopt_encodeIndexVersion(1)` manually for now. At some point version 1 may become the default.

Noteworthy is that the strip-like index order typically is the same or slightly larger than non-strip order in terms of raw index encoding, *however* it results in more compressible vertex data and it's much more compressible with gzip/zstd - after this change, post-deflate most meshes use 3-4 bits per triangle instead of 5-6 with the old order.

Here's an example for a Nefertiti model to illustrate the deltas:

Before: vertex data 19.03 MB, index data 7.48 MB, vertex+index gzip 20.65 MB
After: vertex data 17.64 MB, index data 7.10 MB, vertex+index gzip 17.79 MB

In general, on a large set of geometry-heavy glTF files, the ratios of new data to old data are:
uncompressed index: 0.93x .. 1.08x
uncompressed vertex: 0.93x .. 1.01x
compressed vertex+index: 0.74x .. 1.00x

Some models in this test set have redundant connectivity data (point cloud or disjointed triangles) so on most reasonable models the data does compress better.


zeux:
Not sure about the versions re: API yet, thoughts: https://gist.github.com/zeux/ace5c2372ab05d8e1590a824120f36cf

This might get merged with the existing API to minimize churn and get revised before release...

PR #104, opened 2020-02-12
Vertex packing experiment.

castano:
This is an attempt at improving vertex position quantization by using an iterative least squares optimization procedure. This produces best results at high quantization settings, for example, packing positions in 8 or 10 bits per component. Regular meshes tend to produce much better results than organic ones. For example, in pirate.obj the improvements are minimal, but on teapot.obj I get fairly decent improvement in the mean square error:

non-uniform scale:
UNORM10:   0.00472621 -> 0.00450743  (92%)
UNORM8:    0.07289813 -> 0.06713689  (95%)

uniform scale:
UNORM10:   0.00799596 -> 0.00644280  (81%)
UNORM8:    0.12306081 -> 0.11552908  (94%)

Let me know if this looks interesting at all, or if you have suggestions for improvements. I'd be happy to clean it up for inclusion if you think it's valuable.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -26,6 +26,7 @@ set(SOURCES
     src/vcacheanalyzer.cpp
     src/vcacheoptimizer.cpp
     src/vertexcodec.cpp
+    src/vertexpacker.cpp
     src/vfetchanalyzer.cpp
     src/vfetchoptimizer.cpp
 )
--- a/demo/main.cpp
+++ b/demo/main.cpp
@@ -1024,12 +1024,92 @@ void process(const char* path)
 		processDeinterleaved(path);
 }
 
+void vertexPackingTest(const Mesh* mesh)
+{
+	const float* vertex_ptr = &mesh->vertices[0].px;
+	const size_t vertex_stride = sizeof(Vertex);
+	const size_t vertex_count = mesh->vertices.size();
+
+	float scale[3] = {1, 1, 1};
+	float offset[3] = {0, 0, 0};
+	float errors[2];
+
+	// Compare the error of packing the mesh vertex positions using:
+	printf("Vertex Packing Tests:\n");
+
+	// - Half float naive.
+	meshopt_evaluateHalfQuantizationError(vertex_ptr, vertex_stride, vertex_count, scale, offset, errors);
+	printf("Half Float naive:   %.8f, %.8f\n", errors[0], errors[1]);
+
+	// - Half float centered around the origin.
+	float minv[3], maxv[3];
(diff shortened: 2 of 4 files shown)


zeux:
Interesting! Just to make sure I understand this, this is finding the more optimal quantization scale/offset to minimize the error instead of simply relying on the bounding box - the maximum error is the same, but the average error could be reduced?

I'd need to look at what error reduction gains are possible here on some of the meshes I use in the test set and what the performance characteristics of the code are. Might be interesting for gltfpack.

castano:
Yes, that's the idea. There's not much room for error reduction, but why not pick a better offset/scale if there's one available?

The algorithm is linear in the number of vertices, and proportional to the number of iterations. I currently repeat it 3-6 times, but an obvious improvement would be to stop iterating when the error does not decrease anymore. This is also important to avoid increasing the error, which does happen in some cases.

It may also be interesting to weight the vertices based on the area of the triangles around them, or weight boundary vertices to minimize cracks.

Is your set of test meshes available anywhere?

zeux:
Linear in terms of vertex count is good, I wasn't sure if some more complex minimization with a super-linear complexity is taking place (didn't get a chance to look at the code).

I don't know if I can redistribute some of the meshes I use internally; one possibility is to integrate this to gltfpack as a test (you can plug this into `gltf/stream.cpp` to `prepareQuantizationPosition`, perhaps the easiest option is to concatenate all position streams into one and run the function on the result), and run it on glTF-Sample-Models repository (https://github.com/KhronosGroup/glTF-Sample-Models). You can run it in batch using something like `find ../glTF-Sample-Models/2.0/ -name *.gltf | xargs ./gltfpack -test`. I can do this myself as well in a few days.

Two things I'm mostly curious about:

- Does this help with higher bit count for quantization? gltfpack defaults to 14 position bits, I'd generally expect to use 12-16 per component. Of course sometimes lower bit count might be appropriate but that's pretty dependent on the content.

- Does this help with higher vertex count, or do vertices end up conflicting to the point where optimizing isn't worthwhile?

castano:
This is most useful at lower quantization settings. We used 8-10 bit quantization very aggressively for the mobile version of The Witness, but I guess that's becoming less and less common these days. It's certainly not very useful at 16 bits, and I haven't tried intermediate settings.

I think vertex count should not matter much, but if you are quantizing to the point where multiple adjacent vertices collapse into the same bucket, then this is not going to help much.

I'll integrate this into gltfpack when I get some time. I added the uniform scale mode precisely for that.

zeux:
I think it's unlikely that this experiment will become production-ready in the short term - gltfpack's quantization settings default to 14 position bits when this isn't as important. It's a great experiment but since neither of us have had a chance to look into this in the last year and a half I'll close this to avoid confusion.

PR #106, opened 2020-02-17
gltfpack: Implement vertex filters

zeux:
This change extends MESHOPT_compression extension with support for filters.

In general, it's expected that the attribute codec is given mostly non-redundant data as an input.
However, in glTF context, we're forced to work with the data that glTF specification expects - namely,
unit vectors for normal/tangent data, and unit quaternions for animation data.

These representations are not optimal from the compression point of view. This change introduces the
concept of filters - a post-processing function that runs after the vertex decoding has ran, and transforms the data into ready to use format.

Three filters are implemented here - Oct8 / Oct12 support unit vector storage by decoding octahedral normal vector representation, and Quat12 supports unit quaternion storage by decoding three-component vector and reconstructing the fourth vectors.

All three codecs come with WASM SIMD implementations; SSE2 path isn't implemented yet.

In terms of decompression performance, the implementation is very fast: on latest v8, oct8 runs at 2.02 GB/s, oct12 runs at 3.7 GB/s, quat12 runs at 2.5 GB/s - this is with a baseline vertex decoder running at 1.9 GB/s; timings are from i7-8700K.

Oct8 is selected as a high compression alternative to snorm8 normal/tangent storage; Oct12 is selected as a high precision alternative to snorm16 storage (which is overkill for most cases). Oct8 saves ~5-6 bits from each of normal/tangent compared to snorm8.

Quat12 is selected as a reasonably high precision alternative to snorm16 quaternion storage; due to reconstruction of largest component we achieve a suitably low decoding error of 0.003-0.004 degrees (snorm16 has an error of 0.001-0.002 degrees), and use ~half the bits.


zeux:
.js file before/after this change:

```
$ ls -la js/meshopt_decoder.js*
-rwxrwxrwx 1 zeux zeux 26604 Feb 16 22:20 js/meshopt_decoder.js
-rwxrwxrwx 1 zeux zeux  4823 Feb 16 22:20 js/meshopt_decoder.js.gz
```

=>

```
$ ls -la js/meshopt_decoder.js*
-rwxrwxrwx 1 zeux zeux 32553 Feb 16 21:35 js/meshopt_decoder.js
-rwxrwxrwx 1 zeux zeux  6092 Feb 16 21:35 js/meshopt_decoder.js.gz
```

The increase seems reasonable given the compression benefits.

Example before/after gltfpack sizes:

BrainStem.gltf: before 438 KB (gzip 317 KB), after 370 KB (gzip 256 KB)
Corset.gltf: before 153 KB (gzip 98 KB), after 133 KB (gzip 84 KB)
animation-heavy model: before 3888 KB (gzip 1445 KB), after 2972 KB (gzip 794 KB)...

PR #124, opened 2020-04-16
gltfpack: Fixed missing mime_type

atteneder:
Encoding basisu/ktx2 textures fails if the mime_type is not setup.
basisu command fails if the input file has the extension `.raw` instead of `.png`.

Happened when converting glTF-JSON (with separate texture file) to glTF-JSON (with -tc option).

Tried to upload a sample file, but github throws errors :/

### DIFF
--- a/gltf/write.cpp
+++ b/gltf/write.cpp
@@ -567,6 +567,10 @@ void writeImage(std::string& json, std::vector<BufferView>& views, const cgltf_i
 
 		mime_type = inferMimeType(image.uri);
 	}
+	else if (image.uri)
+	{
+		mime_type = image.mime_type ? image.mime_type : inferMimeType(image.uri);
+	}
 
 	if (!img_data.empty())
 	{


atteneder:
Now it worked:

[example_gltf.zip](https://github.com/zeux/meshoptimizer/files/4488347/example_gltf.zip)

gltf/write.cpp
```diff
@@ -602,6 +602,7 @@ void writeImage(std::string& json, std::vector<BufferView>& views, const cgltf_i
 			if (readFile(full_path.c_str(), img_data))
 			{
 				std::string encoded;
+				mime_type = image.mime_type ? image.mime_type : inferMimeType(image.uri);
```
zeux, line 605: Ah, makes sense - I forgot about this. Can this be moved to the beginning of the function, perhaps as a `else if (image.uri)` after the inferMimeType is being called right now? That will make mime_type more consistently present throughout the rest of the function.

PR #125, opened 2020-04-16
 vertexfilter: Implement NEON versions for all filters

zeux:
Scalar performance:
oct8 231.26 ms (0.13 GB/sec), oct12 231.22 ms (0.26 GB/sec), quat12 179.14 ms (0.33 GB/sec), exp 19.04 ms (3.13 GB/sec)

NEON performance:
oct8 62.71 ms (0.48 GB/sec), oct12 69.98 ms (0.85 GB/sec), quat12 73.33 ms (0.81 GB/sec), exp 19.06 ms (3.13 GB/sec)

Note: the NEON code only compiles on AArch64 at the moment, which will need to be fixed before merging this.


zeux:
Final perf numbers on Amazon A1:

filter: oct8 54.37 ms (0.55 GB/sec), oct12 75.31 ms (0.79 GB/sec), quat12 73.29 ms (0.81 GB/sec), exp 16.13 ms (3.69 GB/sec)

Also this PR improves oct8 for SSE2 (2.2 GB/s => 2.6 GB/s) and adds iPhone 32/64 bit build back so that we can be sure we don't break 32-bit ARM compilation with NEON.

PR #126, opened 2020-04-16
Add 'config=scalar' build flag

abrown:
This allows building codecbench with no SIMD code but still with optimizations enabled. @zeux, there is likely a clearer way to do this so feel free to ask for changes or commit to this branch! The rationale for this is to build an optimized, scalar version of codecbench.

### DIFF
--- a/src/vertexcodec.cpp
+++ b/src/vertexcodec.cpp
@@ -4,6 +4,9 @@
 #include <assert.h>
 #include <string.h>
 
+// Detect SIMD settings.
+#if !defined(MESHOPTIMIZER_NO_SIMD)
+
 #if defined(__ARM_NEON__) || defined(__ARM_NEON)
 #define SIMD_NEON
 #endif
@@ -39,6 +42,8 @@
 #define SIMD_WASM
 #define SIMD_TARGET __attribute__((target("unimplemented-simd128")))
 #endif
+
+#endif // !MESHOPTIMIZER_NO_SIMD
 
 #ifndef SIMD_TARGET
 #define SIMD_TARGET
--- a/Makefile
+++ b/Makefile
@@ -40,6 +40,10 @@ ifeq ($(config),trace)
 	CXXFLAGS+=-DTRACE=2
 endif
 
+ifeq ($(config),scalar)
+	CXXFLAGS+=-O3 -DNDEBUG -DMESHOPTIMIZER_NO_SIMD
+endif
+
 ifeq ($(config),release)
 	CXXFLAGS+=-O3 -DNDEBUG
 endif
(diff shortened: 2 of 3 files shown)


zeux:
Is the idea to be able to profile native binaries with scalar implementations?

abrown:
Yeah, it would be nice to see the speedup caused by enabling SIMD but that only makes sense if the baseline (non-SIMD) is actually optimized.

zeux:
Ok - cool. I was thinking at some point that a configuration option to disable SIMD might be good in case there's any platform or compiler specific issues, so users don't need to edit source code.

It would be good to see a couple changes for this though:

1. The define should be called MESHOPTIMIZER_NO_SIMD so that it can be declared externally as a config option without clashes
2. It would be good to, instead of undef-ing existing macros, just wrap the entire autodetection section at the top of vertexcodec.cpp & vertexfilter.cpp in #ifndef MESHOPTIMIZER_NO_SIMD. That way this define can work in cases where, for example, there's a problem with a platform-provided SIMD intrinsic header so even including it is problematic.

That way you also wouldn't need to change the dispatch logic (I think), just the absence of any SIMD_ macros will be enough.

abrown:
@zeux, updated. Let me know if you want different code style for #if/#endif (I ended up using `define(...)` since that is what I saw in the rest of the code).

src/vertexcodec.cpp
```diff
@@ -1201,7 +1206,9 @@ int meshopt_decodeVertexBuffer(void* destination, size_t vertex_count, size_t ve
 
 	const unsigned char* (*decode)(const unsigned char*, const unsigned char*, unsigned char*, size_t, size_t, unsigned char[256]) = 0;
 
-#if defined(SIMD_SSE) && defined(SIMD_FALLBACK)
+#if defined(MESHOPTIMIZER_NO_SIMD)
```
zeux, line 1209: Is this block still necessary? It seems that if no SIMD_ defines are defined, this should correctly use scalar decoder.

abrown: I didn't see it that clearly: the cpuid check below made me think that perhaps `decodeVertexBlockSimd` could be chosen, even if `MESHOPTIMIZER_NO_SIMD` is set. I can remove this if you are convinced, but perhaps we should add some documentation for the uninitiated (like me)?

zeux: Well, the cpuid check is only performed if SIMD_SSE is defined - if SIMD_SSE isn't defined no SIMD code is compiled (or executed).

abrown: Ah, good point! I wasn't looking very carefully and for some reason I thought it was `SIMD_SSE || SIMD_FALLBACK`. I removed this part and re-pushed.

zeux:
Thanks!

PR #127, opened 2020-04-16
Build Wasm versions of codecbench with debug symbols

abrown:
This will affect `codecbench`'s code size (probably not a high area of concern) but will make the generated Wasm much easier to analyze. Alternately, we could make the `WASM_FLAGS` variable more generic (which would involve some refactoring) and include the `-g` there.


zeux:
Yeah codecbench isn't deployed anywhere so this looks good to me 👍

PR #129, opened 2020-04-18
CMake: make shared libs work again

mosra:
Just a minor buildsystem thing. 

`BUILD_SHARED_LIBS` got renamed to `MESHOPT_BUILD_SHARED_LIBS` by @corporateshark in #91, but the unprefixed variant is what still needs to be set for `add_library()` to work as desired. Without this, enabling `MESHOPT_BUILD_SHARED_LIBS` still produced an `*.a` file and additionally made CMake complain about visibility presets being set for static libs.

By the way, I packaged the latest release for ArchLinux and will be maintaining it for the foreseeable future (https://aur.archlinux.org/packages/meshoptimizer/), would it make sense to mention it in the README?

Thank you!

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -16,6 +16,11 @@ option(MESHOPT_BUILD_TOOLS "Build tools" OFF)
 option(MESHOPT_BUILD_SHARED_LIBS "Build shared libraries" OFF)
 set(MESHOPT_BUILD_TOOLS_GLFW_FOLDER_NAME "" CACHE STRING "Custom folder to look for GLFW")
 
+# add_library() needs BUILD_SHARED_LIBS set to decide between static and
+# shared. This also sets it back to OFF in case parent project has
+# BUILD_SHARED_LIBS but wants MESHOPT_BUILD_SHARED_LIBS to be OFF.
+set(BUILD_SHARED_LIBS ${MESHOPT_BUILD_SHARED_LIBS})
+
 set(SOURCES
     src/meshoptimizer.h
     src/allocator.cpp


zeux:
Ah, good point! Since this option is a standard CMake option I am wondering if we. need a MESHOPT prefixed one, or if just using unprefixed option is enough?

zeux:
Oh, also happy to mention the package. Please let me know if any other changes can be made to make packaging easier for meshoptimizer/gltfpack.

mosra:
> I am wondering if we need a MESHOPT prefixed one

I could imagine a use case where you might want the two separate -- for [magnum](https://github.com/mosra/magnum) I'm integrating meshoptimizer as a plugin (which is shared/dynamic), but meshoptimizer itself could be static to have the plugin binary self-contained. With a single global option could be of course done also with for example

```cmake
set(BUILD_SHARED_LIBS OFF)
add_subdirectory(meshoptimizer)
set(BUILD_SHARED_LIBS ON)
```

but I think having two separate options is a cleaner variant. I'm fine with either, so let me know if I should change to just a single global one.

> packaging

I was positively surprised how smooth this was, actually :) I think it could be useful to have a separate option for enabling `gltfpack` without `lodviewer` (plus lodviewer fails to compile for me currently). I'll submit a PR with that next.

zeux:
I see - separate options make sense. `lodviewer` is meant as an internal debugging tool and shouldn't be distributed anyway, it makes sense to build it separately.

zeux:
For lodviewer I think it might make sense to not build it by default when _TOOLS are being built; at some point it might be removed from the repository, or lose glfw dependency, etc. so I don't want to commit to it being everpresent.

mosra:
I'd personally add a dedicated `MESHOPT_BUILD_GLTFPACK` option and leave `_TOOLS` for lodviewer and potential other internal debugging tools. 

(My first experience when building the project was "but where's gltfpack?" and it wasn't immediately obvious that I need to enable the `_TOOLS` option to get it.)

zeux:
> I'd personally add a dedicated MESHOPT_BUILD_GLTFPACK option

This sounds good to me as well.

PR #130, opened 2020-04-18
Fix build on old Apple Clang versions

mosra:
*(I have to admit my monster [build matrix](https://magnum.graphics/build-status/) contains a bunch of outdated systems.)*

On Apple platforms, `__clang_major__` follows a different versioning scheme and version 8 is actually vanilla Clang 3.9, which made the build fail due to `__builtin_rotateleft64` not being defined. Extended the condition to check for version 11 instead if the compiler is Apple Clang. [Version mapping table on Wikipedia](https://en.wikipedia.org/wiki/Xcode#Toolchain_versions) for reference.

Also added an [.editorconfig](https://editorconfig.org) so editors implicitly know to indent with tabs plus a bunch of other good defaults.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,8 @@
+# See https://editorconfig.org/ for more info
+
+[*]
+charset = utf-8
+indent_style = tab
+indent_size = 4
+trim_trailing_whitespace = true
+insert_final_newline = true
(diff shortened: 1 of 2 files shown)


zeux:
... wow. I'll learn to never trust clang major again :) (I thought of using __has_builtin, I guess I'll refactor using that at some point). Thanks!

PR #131, opened 2020-04-18
Add a dedicated option to build gltfpack

mosra:
As discussed in #129, this adds a dedicated `MESHOPT_BUILD_GLTFPACK` option.

It's off by default, and I didn't handle backwards compatibility in any way -- maybe it could make sense to make it implicitly `ON`? Except if majority of the users need only the library, then building it always would be probably unnecessary.

I also made a bunch of fixes to make `lodviewer` compile at least halfway on Linux/macOS (adding a missing include, silencing `-Werror=unused-parameter` and correctly linking to glfw also if it's found in system locations). It still fails to link due to unresolved references to GL symbols, and for that I'd need to replace the linker pragma with `find_package(OpenGL)` etc. Since you said it's an internal tool anyway, I'm not sure if that's worth the effort :) I can also drop the second commit altogether if you're not interested in these changes.


zeux:
> It's off by default, and I didn't handle backwards compatibility in any way -- maybe it could make sense to make it implicitly ON? Except if majority of the users need only the library, then building it always would be probably unnecessary.

It's probably fine to keep it off - if it's on then applications that depend on the library will automatically get gltfpack as part of the target list which is probably undesirable.

And yeah I don't think fixing lodviewer for non-Windows systems is high value...

mosra:
> if it's on then applications that depend on the library will automatically get gltfpack as part of the target list

Right, good point to keep it off then.

Discarding the lodviewer changes altogether, then?

zeux:
I merged the lodviewer changes as well since they are benign. Thanks!

mosra:
Thanks for the ultra-fast feedback! :+1:

PR #143, opened 2020-05-13
gltfpack: Added -km (keep unused materials) command line option

atteneder:


zeux:
Thanks! Just a minor question - for the use case in question, are the materials named? There was a request at some point to keep named materials in tact without merging them; it would be nice to change the behavior here later to only keep named materials and prevent merging - this doesn't have to happen as part of this change, I just want to know if it's safe to do it later.

atteneder:
In my use case the materials are named, yes. But the name isn't relevant, as there is only one material per glTF and I access simply it via index 0.
I guess keeping name materials only would work for me as well.

zeux:
Ok - I'm going to merge this as is but I'd like to rework this later to be name-based.

The motivation is as follows: as a general rule of thumb, right now gltfpack treats names as not-semantically-meaningful, but there are obvious use cases where you'd want to manipulate the objects by name externally. E.g. gltfpack by default merges equivalent materials (which is really important for some use cases), but an application that wants to go beyond just displaying the content and have a way to change materials should have a way to do so - that ideally doesn't involve maintaining indices because indices are fragile.

So it would be good to have a general behavior where:
- gltfpack by default treats various objects (nodes, materials, meshes) as having structural equivalence and is free to merge / remove / create them as applicable
- For each object type, there's a command line switch that allows to say "things that have a name are meaningful to me".

atteneder:
Yes, that makes sense.
Thanks for merging!

PR #153, opened 2020-06-16
Update gltfpack and loaders to EXT_meshopt_compression

zeux:
This reflects changes from https://github.com/KhronosGroup/glTF/pull/1830 where the extension is now EXT-prefixed, and the enums are using strings instead of integers.

meshopt_decoder.js supports both index-based and string-based filter/mode parameters for now; the loaders were updated with string-based enums, although it's easy to support both. gltfpack only supports the EXT_ extension now.

This PR will stay open for a bit; I'd like to get an indication that the names proposed in the glTF spec won't drift, so to reduce churn this will be merged once the glTF extension is closer to being merged.


zeux:
This change is now fully backwards-compatible in that the updated GLTFLoader.js will happily load MESHOPT_compression assets as well as EXT_meshopt_compression, which should make the upgrade path a bit easier.

The separate Three.JS/Babylon.JS loader modules are just for the EXT_ version however; in the future EXT_ version will be the only one supported.

zeux:
Alright I think I'm going to just merge this; there might be some churn if the glTF extension discussion goes in the direction of changing some of the names but I'd like to make some other changes to loaders/gltfpack here and it's not clear how long this can keep waiting.

PR #159, opened 2020-06-25
demo: Update to three.js r118 and start using plugins

zeux:
Instead of using demo/GLTFLoader.js we now use the plugin infrastructure
added in r118 to load files with EXT_meshopt_compression.

Since three.js is going to deprecate non-module-based examples, the
THREE extension is using `export` as well to conform to the overall style.

Unlike demo/GLTFLoader.js, this method doesn't support mesh instancing
or embedded Basis textures, but this hopefully will be added to
three.js in the future, and for now demo/GLTFLoader.js is still here.

### DIFF
--- a/demo/THREE.EXT_meshopt_compression.js
+++ b/demo/THREE.EXT_meshopt_compression.js
@@ -38,3 +38,6 @@ var EXT_meshopt_compression = (function () {
 
     return EXT_meshopt_compression;
 }());
+
+/* three.js uses JS modules exclusively since r124 */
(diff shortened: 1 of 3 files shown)


zeux:
three.js changes to watch for GLTFLoader.js to become fully obsolete:

https://github.com/mrdoob/three.js/pull/18913 EXT_mesh_gpu_instancing
https://github.com/mrdoob/three.js/pull/18490 KTX2Loader (not sufficient by itself, also needs GLTFLoader integration for mimeType)

PR #167, opened 2020-07-16
Mention the Vcpkg package in README

mosra:
Following ArchLinux, meshoptimizer is now also available in Vcpkg (https://github.com/microsoft/vcpkg/pull/12211).


zeux:
Thanks!

PR #188, opened 2020-10-10
gltfpack: Preserve scenes during processing

### LINKED ISSUE #184: Process multiple scenes without merging them
I really love gltfpack. The speed and quality of code is superb. But I wish there was a bit more control over certain aspects of it. One of them is ability to process multiple scenes. Right now gltfpack outputs a single scene. So i have to use `gltf-transform` to preprocess file to turn scenes into root nodes.

Another semi-related thing would be super lovely to support multiple input files and glue them together as different scenes. Right now i also do that with `gltf-transform`, but it all takes a lot of plumbing and hacks

zeux:
Before this change gltfack wasn't scene aware, and would process all
nodes regardless of which scene they come from.

With this change, we now output the same number of scenes that the input
file contained. When merging meshes and converting nodes to mesh instances,
the process is limited to cases where the mesh is not used in multiple
scenes.

Fixes #184

### DIFF
--- a/gltf/gltfpack.cpp
+++ b/gltf/gltfpack.cpp
@@ -231,6 +231,7 @@ static void process(cgltf_data* data, const char* input_path, const char* output
 
 	std::vector<NodeInfo> nodes(data->nodes_count);
 
+	markScenes(data, nodes);
 	markAnimated(data, nodes, animations);
 
 	for (size_t i = 0; i < meshes.size(); ++i)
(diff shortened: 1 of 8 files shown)


Inviz:
Thank you very much, master zeux! 

May I ask about your plans to release the current `gltfpack` on npm?

zeux:
0.15 should release later this month.

PR #189, opened 2020-10-12
gltfpack: Provide a CLI and library version of gltfpack

### LINKED ISSUE #164: Provide gltfpack as a js module and also a cli tool.
Can there be a module for gltfpack and also alongside the cli, using the https://docs.npmjs.com/files/package.json#bin .

Having a module would make easier for us folks here at [Scapic](https://scapic.com/) to integrate the same. We do have a very naive implementation right now.

Would love to know if this fits in your roadmap and would love to contribute if there is any help needed.

zeux:
This PR reworks gltfpack Wasm module as a reactor that provides a function to run packing process.
It can be ran repeatedly with different inputs.

Instead of using the global argc/argv mechanism we simply pass the relevant data to the pack function; this allows to omit the WASI argc/argv implementation.

Instead of using getenv mechanism we replace the command on the JS side for the execute() function.

The WASI layer is implemented in a node.js-agnostic way, with file I/O using typed buffers (Uint8Array); cli.js uses node.js Buffer objects to implement these but an arbitrary implementation should be possible.

Here's a primitive example of running this in the web browser (without file I/O implemented):

```
gltfpack.init(fetch('./library.wasm').then(resp => resp.arrayBuffer()));
gltfpack.pack([], {}).catch(console.log);
```

Fixes #164.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -86,6 +86,25 @@ jobs:
       env:
         VALIDATOR: https://github.com/KhronosGroup/glTF-Validator/releases/download/2.0.0-dev.3.2/gltf_validator-2.0.0-dev.3.2-linux64.tar.xz
 
+  gltfpackjs:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v1
+    - uses: actions/setup-node@v1
+      with:
+        node-version: '14.x'
+    - name: install wasi
+      run: |
+        curl -sL https://github.com/WebAssembly/wasi-sdk/releases/download/wasi-sdk-$VERSION/wasi-sdk-$VERSION.0-linux.tar.gz | tar xz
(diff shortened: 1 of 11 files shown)


zeux:
I think this is done now. Here's a minimal node.js example (that omits support for texture compression):

```
#!/usr/bin/env node
var gltfpack = require('./library.js');
var fs = require('fs');

gltfpack.init(fs.readFileSync(__dirname + '/library.wasm'));
gltfpack
	.pack(process.argv.slice(2), { read: fs.readFileSync, write: fs.writeFileSync })
	.then(log => process.stdout.write(log), process.exit(0))
	.catch(err => process.stderr.write(err.message), process.exit(1));
```

zeux:
I think this is done now. Here's a minimal example for node.js:

```
#!/usr/bin/env node
var gltfpack = require('gltfpack');
var fs = require('fs');

gltfpack
	.pack(process.argv.slice(2), { read: fs.readFileSync, write: fs.writeFileSync })
	.then(log => process.stdout.write(log), process.exit(0))
	.catch(err => process.stderr.write(err.message), process.exit(1));
```

PR #190, opened 2020-10-17
gltfpack: More precise and explicit controls over compression/quantization

zeux:
This change replaces -c and -cc with `-c N` to specify the compression level (0 = no compression, 1 = basic compression, 2 = thorough compression) and -noq with `-q N` to specify the quantization level (0 = no quantization, 1 = quantize everything that can be quantized without extra extensions, 2 = quantize positions and texture coordinates as well).

Although the issue is that -q 0 actually still quantizes colors and joint weights so I'm not sure this is a good way to specify these... Ideally this would be changed so that -q 0 doesn't do that before merging, although that changes -noq behavior. Alternatively maybe all of this is a bad idea and we should instead just have binary compression & quantization options, where -noq only disables quantization on position and texture coordinates...

### DIFF
--- a/gltf/gltfpack.cpp
+++ b/gltf/gltfpack.cpp
@@ -649,7 +649,7 @@ static void process(cgltf_data* data, const char* input_path, const char* output
 
 	const ExtensionInfo extensions[] = {
 	    {"KHR_mesh_quantization", settings.quantize, true},
-	    {"EXT_meshopt_compression", settings.compress, !settings.fallback},
+	    {"EXT_meshopt_compression", settings.compress > 0, !settings.fallback},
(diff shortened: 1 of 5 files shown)


zeux:
After mulling this over for a bit more I'm going to close this for now. `-q 1` is a great idea except that it still requires the quantization extension (without changing the node structure, but still) for normal/tangent data, and produces files that are quite a bit larger vs -q 2, so it's not a great option; `-c` changes are also a bit of a double-edged sword, since e.g. `-cc` changes the triangle order from the baseline, which means that the decoded results aren't binary equivalent anymore. There's no clear cut way to decide here, so I'm going to err on the side of keeping the status quo, and we can still revisit this before 1.0.

Issue #198, opened 2020-12-04
Keeping extras from VRM


fire:
Hello,

I have a vrm file. It's a modification of a glb file.

When I rename it to a glb, and then do .\gltfpack.exe -ke -kn -km -ke -i '.\Short Haired Girl Black Dress Less Poly 3.glb'  -o 'Short Haired Girl Black Dress Less Poly 4.glb'.

VRM importers fails to load it as a vrm.

I have tried the `-ke` option.

Any suggestions?

[Short Haired Girl Black Dress Less Poly 3.zip](https://github.com/zeux/meshoptimizer/files/5643587/Short.Haired.Girl.Black.Dress.Less.Poly.3.zip)

All Rights Reserved. Ask me if you want license details.

https://github.com/saturday06/VRM_IMPORTER_for_Blender is an blender addon for VRM.

avaer:
It would be useful to us too if the VRM extensions were not removed during the meshopt process. 👍

fire:
This is my attempt to patch meshoptimizer.

https://github.com/fire/meshoptimizer/pull/1

Looking for suggestions.

```
git clone https://github.com/fire/meshoptimizer.git -b vrm 
cd meshoptimizer
mkdir build
cd build
cmake -DMESHOPT_BUILD_GLTFPACK=ON ..
make
wget https://github.com/Miraikomachi/MiraikomachiVRM/raw/master/Miraikomachi.vrm
./gltfpack -si 0.1 -noq -ke -ke -i "Miraikomachi.vrm" -o "Miraikomachi Optimize.vrm"
# Open in a vrm importer.
```

zeux:
I'm not familiar with VRM but if it's just about preserving the top-level extension node "VRM", it should be simple. The change above might be too broad though, as it would preserve some top-level extensions where extra work needs to be performed to make the content work, e.g. KHR_xmp.

My suggestion thus would be to simply detect when "VRM" is present in the extension list (regardless of input options) and, if so, output just that extension blob, using `appendJson` which performs JSON minification.

Feel free to submit a PR along these lines. I don't have a way to test this so testing suggestions would be welcome.

fire:
```
BEFORE 7.6M AliciaSolid_vrm-0.51.vrm
AFTER 5.5M   AliciaSolid_vrm-0.51-opt.vrm
```

From https://github.com/vrm-c/UniVRM/tree/master/Tests/Models/Alicia_vrm-0.51

[...]

PR #199, opened 2020-12-05
Support for vrm.

fire:
Use a vrm file extension to trigger processing.

See #198 for test cases and examples.

The command to test is now:

`./gltfpack -i "Miraikomachi.vrm" -o "Miraikomachi Opt.VRM" `

### DIFF
--- a/gltf/gltfpack.cpp
+++ b/gltf/gltfpack.cpp
@@ -667,7 +667,7 @@ static void process(cgltf_data* data, const char* input_path, const char* output
 	const ExtensionInfo extensions[] = {
 	    {"KHR_mesh_quantization", settings.quantize, true},
 	    {"EXT_meshopt_compression", settings.compress, !settings.fallback},
-	    {"KHR_texture_transform", settings.quantize && !json_textures.empty(), false},
+	    {"KHR_texture_transform", (settings.quantize && !json_textures.empty()) || (settings.vrm && !json_textures.empty()), false},
 	    {"KHR_materials_pbrSpecularGlossiness", ext_pbr_specular_glossiness, false},
 	    {"KHR_materials_clearcoat", ext_clearcoat, false},
 	    {"KHR_materials_transmission", ext_transmission, false},
--- a/gltf/gltfpack.cpp
+++ b/gltf/gltfpack.cpp
@@ -900,6 +900,7 @@ int gltfpack(const char* input, const char* output, const char* report, Settings
 	{
 		settings.vrm = true;
 		settings.keep_extras = true;
+		settings.quantize = false;
 	}
 
 	std::string json, bin, fallback;
--- a/gltf/gltfpack.cpp
+++ b/gltf/gltfpack.cpp
@@ -442,7 +442,7 @@ static void process(cgltf_data* data, const char* input_path, const char* output
 
 		comma(json_meshes);
 		append(json_meshes, "{");
-		if (mesh.name)
+		if (!mesh.name.empty())
 		{
 			append(json_meshes, "\"name\":\"");
 			append(json_meshes, mesh.name);
--- a/gltf/gltfpack.cpp
+++ b/gltf/gltfpack.cpp
@@ -899,10 +899,6 @@ int gltfpack(const char* input, const char* output, const char* report, Settings
 	{
 		settings.vrm = true;
 		settings.keep_extras = true;
-		settings.keep_materials = true;
-		settings.mesh_merge = false;
-		settings.keep_nodes = true;
-		settings.quantize = false;
 	}
 
 	std::string json, bin, fallback;
--- a/gltf/gltfpack.cpp
+++ b/gltf/gltfpack.cpp
@@ -357,6 +357,7 @@ static void process(cgltf_data* data, const char* input_path, const char* output
 	std::string json_accessors;
 	std::string json_meshes;
 	std::string json_nodes;
+	std::string json_extensions;
 	std::string json_skins;
 	std::vector<std::string> json_roots(data->scenes_count);
 	std::string json_animations;
@@ -440,7 +441,15 @@ static void process(cgltf_data* data, const char* input_path, const char* output
 		const Mesh& mesh = meshes[i];
 
 		comma(json_meshes);
-		append(json_meshes, "{\"primitives\":[");
+		append(json_meshes, "{");
+		if (mesh.name)
+		{
+			append(json_meshes, "\"name\":\"");
+			append(json_meshes, mesh.name);
+			append(json_meshes, "\"");
+			comma(json_meshes);
+		}
+		append(json_meshes, "\"primitives\":[");
 
 		size_t pi = i;
 		for (; pi < meshes.size(); ++pi)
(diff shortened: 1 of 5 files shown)


fire:
How do I read those errors in the cicd?

fire:
@zeux I wasn't able to recreate the errors. Do you know how that test works?

zeux:
The test validates that packing any model from glTF-Sample-Models repository produces a valid glTF file as validated by glTF-Validator. glTF-Sample-Models/2.0/Cameras/glTF/Cameras.gltf is the specific file on which the process fails with this change, you should be able to get the source file from glTF-Sample-Models Khronos repository.

fire:
Green!

gltf/parsegltf.cpp
```diff
@@ -165,6 +175,8 @@ static void parseMeshesGltf(cgltf_data* data, std::vector<Mesh>& meshes, std::ve
 			meshes.push_back(Mesh());
 			Mesh& result = meshes.back();
 
+			result.name = mesh.name;
```
fire, line 178: Might be a problem with variable lifetime

gltf/gltfpack.cpp
```diff
@@ -488,6 +498,11 @@ static void process(cgltf_data* data, const char* input_path, const char* output
 				append(json_meshes, ",\"indices\":");
 				append(json_meshes, index_accr);
 			}
+			if (settings.keep_extras && !prim.extras.empty())
```
zeux, line 501: Why is this necessary? AFAICT the only "extras" node in the .VRM file from the issue is on the mesh, not on primitive, and that contains targetNames which should already be preserved by gltfpack.

gltf/gltfpack.cpp
```diff
@@ -661,10 +678,39 @@ static void process(cgltf_data* data, const char* input_path, const char* output
...
+	append(json_extensions, "\"extensions\":{");
+	if (settings.vrm)
+	{
+
+		std::string extension_vrm;
+		extension_vrm.assign("VRM");
+		for (size_t i = 0; i < data->data_extensions_count; ++i)
+		{
+			cgltf_extension extension = data->data_extensions[i];
+			if (!extension.name)
```
zeux, line 696: This & above & below can be replaced with `if (strcmp(extension.name, "VRM") == 0)` (extension.name is guaranteed to be non-null)

gltf/gltfpack.cpp
```diff
@@ -676,6 +722,8 @@ static void process(cgltf_data* data, const char* input_path, const char* output
 	writeArray(json, "meshes", json_meshes);
 	writeArray(json, "skins", json_skins);
 	writeArray(json, "animations", json_animations);
+	comma(json);
+	append(json, json_extensions);
```
zeux, line 726: This will always emit "extensions" and also causes a conflict with top-level KHR_lights_punctual extension since there's now going to be two extension blocks for files that use that.

Instead what should happen is:
- `json_extensions` should be initially empty
- `json_lights` should be appended to it, if necessary
- VRM data should be appended to it, if necessary
- if `json_extensions` isn't empty when forming the final JSON blob, we should write out `"extensions":{` followed by json_extensions contents followed by `}`

gltf/gltfpack.cpp
```diff
@@ -661,10 +678,39 @@ static void process(cgltf_data* data, const char* input_path, const char* output
...
+			std::string extension_tested;
+			extension_tested.assign(extension.name);
+			if (extension_vrm != extension_tested)
+			{
+				break;
+			}
+			append(json_extensions, "\"");
+			append(json_extensions, extension.name);
+			append(json_extensions, "\":");
+			append(json_extensions, extension.data);
```
zeux, line 709: This should use `appendJson` to minify the output

gltf/gltfpack.cpp
```diff
@@ -790,8 +838,12 @@ int gltfpack(const char* input, const char* output, const char* report, Settings
 
 	const char* iext = strrchr(input, '.');
 	const char* oext = output ? strrchr(output, '.') : NULL;
+	
+	if (iext && (strcmp(iext, ".vrm") == 0 || strcmp(iext, ".VRM") == 0)) {
```
zeux, line 842: Please use clang-format on this code, as this bracing style is inconsistent with the coding guidelines, see https://github.com/zeux/meshoptimizer/blob/master/CONTRIBUTING.md

gltf/gltfpack.cpp
```diff
@@ -843,6 +895,16 @@ int gltfpack(const char* input, const char* output, const char* report, Settings
 		settings.texture_embed = true;
 	}
 
+	if (oext && (strcmp(oext, ".vrm") == 0 || strcmp(oext, ".VRM") == 0))
+	{
+		settings.vrm = true;
+		settings.keep_extras = true;
```
zeux, line 901: I would prefer to omit these, esp. `quantize` but preferably others. It's not clear to me why these need to be present, and if a user wants to apply these optimizations to a VRM file there's not going to be a way to do this without adding more kinds of command-line options, if `vrm` extension acts as an override.

gltf/gltfpack.cpp
```diff
@@ -896,7 +958,7 @@ int gltfpack(const char* input, const char* output, const char* report, Settings
 			return 4;
 		}
 	}
-	else if (oext && (strcmp(oext, ".glb") == 0 || strcmp(oext, ".GLB") == 0))
+	else if (oext && (strcmp(oext, ".vrm") == 0 || strcmp(oext, ".VRM") == 0 || strcmp(oext, ".glb") == 0 || strcmp(oext, ".GLB") == 0))
```
zeux, line 961: Should this also be added to the condition above that sets `texture_embed`? Unclear what the de-facto rules are for VRM files, e.g. whether they are supposed to keep textures separately or not.

gltf/gltfpack.h
```diff
@@ -39,6 +39,7 @@ struct Transform
 
 struct Mesh
 {
+	const char* name;
```
zeux, line 42: `std::string` might be better.

gltf/write.cpp
```diff
@@ -739,6 +739,14 @@ static std::string decodeUri(const char* uri)
 
 void writeImage(std::string& json, std::vector<BufferView>& views, const cgltf_image& image, const ImageInfo& info, size_t index, const char* input_path, const char* output_path, const Settings& settings)
 {
+	if (image.name)
```
zeux, line 742: Is preserving the name of the image object necessary? It doesn't carry semantical meaning, and gltfpack right now only preserves names for objects that are likely to be used by-name externally, specifically nodes, materials and animations.

gltf/write.cpp
```diff
@@ -1320,7 +1328,8 @@ void writeAnimation(std::string& json, std::vector<BufferView>& views, std::stri
 
 		comma(json_samplers);
 		append(json_samplers, "{\"input\":");
-		append(json_samplers, range ? range_accr : track.constant ? pose_accr : time_accr);
+		append(json_samplers, range ? range_accr : track.constant ? pose_accr
```
zeux, line 1331: Was this change intended?

fire:
Thank you for the review. I will be delayed until the weekend to submit the changes.

gltf/write.cpp
fire: Not intended.

gltf/write.cpp
fire: I need to check if this is required by the internal material definition called MTOON.

fire:
I slept thinking about this optimization, because Godot Engine added autolod, I don't have a requirement for this. This VRM pr is salvageable.

arpu:
still interessded in vrm optimation

@fire  can you point me to how Godot Engine autolod works?

fire:
1. Godot load gltf2
2. For every imported mesh run mesh optimizer with the error returned
3. (Would like tesselate and the decimate option) because simpify gets stucks
4. Implement vrm

I believe that mesh optimizer works under wasm, so you can follow the same flow.

PR #200, opened 2020-12-09
Bump GLTF Validator version

zeux:


zeux:
Nice, this found a bug with clearcoat extension handling!

PR #208, opened 2020-12-22
Change .clang-format ColumnLimit to 120

zeux:
clang-format 11 introduces bugs around ColumnLimit: 0 that result in
objectively ugly formatting.

This is an experimental PR of what would happen if we were to enforce a
finite column limit; 120 characters is picked because my text editor
setup fits 2 120-wide views side by side ;)

This can't be merged as is, as it hits other issues with clang-format
around comment reflow and array initializers, but might be reasonable
to switch to once these get addressed.

Will probably require further tuning of .clang-format.


zeux:
Still doesn't spark joy :( Will need to experiment more...

PR #211, opened 2020-12-24
simplify: Restrict edge collapses to avoid triangle flipping

### LINKED ISSUE #86: Simplification runs, but the result has many defects/issues
I am using meshoptimizer's simplification methods to simplify 3D scans w/ have been triangulated using marching cubes. The meshes are far more dense than they need to be, so I am attempting to simplify them down to 20% of their original density.

Unfortunately the result is unusable and full of holes and other defects. Is this normal?

Here is an example:

**Input Mesh**

image

**Simplified Result**

image

I am using MeshLab to visualize the result. The meshes are exported to STL format.

The mesh format I am using internally in my software is that of the VCG library, but I think I am appropriately converting the data such that meshoptimizer's simplification methods can operate correctly.

Here is how I am converting the data:

```
const MeshVert* __restrict root = input.vert.data();
std::vector<unsigned> remap;
std::vector<vcg::Point3f> vertices;

for (const auto& v : input.vert) 
    vertices.emplace_back(v.P());

for (const auto& f : input.face) {
    remap.emplace_back(unsigned((MeshVert*)(f.cV(0)) - (MeshVert*)root));
    remap.emplace_back(unsigned((MeshVert*)(f.cV(1)) - (MeshVert*)root));
    remap.emplace_back(unsigned((MeshVert*)(f.cV(2)) - (MeshVert*)root));
}

const size_t index_count = remap.size();
std::vector<unsigned> simplified(index_count);
size_t target_count = simplification_factor * index_count;
float target_error = 1e-3f;

if (restricted)
    simplified.resize(meshopt_simplify(&simplified[0], remap.data(), index_count, (const float* __restrict)&vertices[0], vertices.size(), sizeof(vcg::Point3f), target_count, target_error));
else
    simplified.resize(meshopt_simplifySloppy(&simplified[0], remap.data(), index_count, (const float* __restrict)&vertices[0], vertices.size(), sizeof(vcg::Point3f), target_count));
```

I'm a little confused by the concept of "Index" in this mesh processing library. Are the indexes here referring to triplets of offsets from Vertex_Ptr* + 0 which represent each face/triangle?

The VCG mesh has a vector of vertices, each represented by a Point3f type (the internal data of which is 3 floats - X, Y, and Z). The faces contain 3 vertex pointers. I compute the offset/index from these pointers by subtracting the start of the vertex array (root).

Is this the correct approach? Am I doing something wrong? The result looks *reasonable*, but not entirely correct. Perhaps this is just the nature of decimation?

### LINKED ISSUE #206: Issues with flat geometry in simplify
OK After a lot of testing, I am realizing that simplify has issues with flat geometry. I suppose the error is detected relative to the distance from the removed point to its new position, or something similar. In geometry with volume, this makes sense, but in flat geometry this appears to cause issues. Here is an example of such thing in a part of the Bistro from Crytek. 

https://user-images.githubusercontent.com/6265307/102661871-6f989780-415c-11eb-8968-d525358b9b0a.mp4

Attached also the geometry in gltf format
[casito.zip](https://github.com/zeux/meshoptimizer/files/5718066/casito.zip)

IMO the problem here is that the algorithm needs to properly consider also the distance from the point to the unconnected edges and use it as error.

image

zeux:
Right now simplifier does not restrict edge collapses beyond
topology/quadric errors, which in some cases results in triangles
getting flipped.

This is a well known problem and a well known solution is to check
whether the triangles are flipped explicitly during collapse.

Unfortunately, the simplifier currently doesn't maintain enough data to
be able to perform this analysis; this change adds code to maintain a
second adjacency structure (that identifies triangles during collapse)
and uses it to validate collapses right before they happen.

~The current implementation seems to resolve existing issues with the
flipping but results in substantially worse simplification performance
(4-5x slowdown, most of the work seems to be in hasTriangleFlips). This
will *not* be merged in this state; I'm going to try a couple of different
approaches to cache and/or otherwise adjust the computation to be faster.~

This implementation makes simplification 20-30% slower, but we get improved
resulting meshes and a new adjacency structure that's maintained throughout
collapse passes which may prove beneficial in the future.

Before this change:
image

After this change:
image

Fixes #86
Fixes #206

### DIFF
--- a/demo/main.cpp
+++ b/demo/main.cpp
@@ -456,10 +456,6 @@ void simplify(const Mesh& mesh, float threshold = 0.2f)
 
 	double end = timestamp();
 
-#if TRACE
-	dumpObj(lod, /* recomputeNormals= */ true);
-#endif
-
 	printf("%-9s: %d triangles => %d triangles in %.2f msec\n",
 	       "Simplify",
 	       int(mesh.indices.size() / 3), int(lod.indices.size() / 3), (end - start) * 1000);
--- a/demo/main.cpp
+++ b/demo/main.cpp
@@ -138,6 +138,69 @@ Mesh parseObj(const char* path, double& reindex)
 	return result;
 }
 
+void dumpObj(const Mesh& mesh, bool recomputeNormals = false)
+{
+	std::vector<float> normals;
+
+	if (recomputeNormals)
+	{
+		normals.resize(mesh.vertices.size() * 3);
+
+		for (size_t i = 0; i < mesh.indices.size(); i += 3)
+		{
+			unsigned int a = mesh.indices[i], b = mesh.indices[i + 1], c = mesh.indices[i + 2];
+
+			const Vertex& va = mesh.vertices[a];
+			const Vertex& vb = mesh.vertices[b];
+			const Vertex& vc = mesh.vertices[c];
+
+			float nx = (vb.py - va.py) * (vc.pz - va.pz) - (vb.pz - va.pz) * (vc.py - va.py);
+			float ny = (vb.pz - va.pz) * (vc.px - va.px) - (vb.px - va.px) * (vc.pz - va.pz);
+			float nz = (vb.px - va.px) * (vc.py - va.py) - (vb.py - va.py) * (vc.px - va.px);
+
+			for (int k = 0; k < 3; ++k)
+			{
+				unsigned int index = mesh.indices[i + k];
+
+				normals[index * 3 + 0] += nx;
+				normals[index * 3 + 1] += ny;
+				normals[index * 3 + 2] += nz;
+			}
+		}
+	}
+
+	for (size_t i = 0; i < mesh.vertices.size(); ++i)
+	{
+		const Vertex& v = mesh.vertices[i];
+
+		float nx = v.nx, ny = v.ny, nz = v.nz;
+
+		if (recomputeNormals)
+		{
+			nx = normals[i * 3 + 0];
+			ny = normals[i * 3 + 1];
+			nz = normals[i * 3 + 2];
+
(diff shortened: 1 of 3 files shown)


zeux:
With some further improvements that went in the performance noticeably improved; this is now running at ~0.5x the performance of the original simplifier, which is more in line with what I expected originally. I'm hoping to improve this further though, the poor performance of mode 1 (which I would like to settle on) is due to the fact that the simplifier ends up doing many more passes on some inputs due to how the pass error bound is configured, which seems solvable.

reduz:
This is amazing, thanks hugely! Will try it asap in Godot.

zeux:
Looks like this doesn't really solve #86 unfortunately, or at least doesn't solve the two model examples attached to that issue. I'm not sure I understand why; I tested a couple modifications of this change but they also didn't work, which leads me to believe that there's something I'm missing here; this will take a bit of time to debug.

zeux:
Ok, the latest update here does fix the problem in #86 - I've been misled by MeshLab's default normal computation mode using some computations that aren't handling thin triangles well, and on those models the tessellation of the simplified result could use some regularization. Having said that, the latest update makes simplification slower again for complex reasons :( so I will need to look into performance again...

zeux:
I'm finally done with this 😅 this was much more difficult than I expected. The simplifier is ~20-30% slower than it was before this change which makes me feel much better. I'm going to clean up the tracing code a bit before submitting but this should be ready for testing, as I don't think the logic is going to change further.

happyfrank:
I tested a small model and it still seems to have this problem -.-

zeux:
Please add this model to one of the two open issues on this.

happyfrank:
Thanks that's all right. ^v^

PR #215, opened 2020-12-28
Removed ossfuzz make rule

AdamKorcz:
The rule is not being used as it is now. It may be brought back at a later time, but it is being removed now to avoid having unused code.


zeux:
Just out of curiosity, I've seen the change in the oss-fuzz that uses CMake to build the library - two questions on that note:

1) Would it be easier to use the .a library that CMake build uses? Not clear why you need to use `ar` to rebuild it from .o artefacts
2) Does oss-fuzz setup CMake environment automatically to introduce -fsanitize flags? Asking because CMake setup doesn't have a fuzz mode but maybe this magically happens

AdamKorcz:
> Would it be easier to use the .a library that CMake build produces?

Yes, it probably will be. I will try it out. Thank you for pointing it out.

> Does oss-fuzz setup CMake environment automatically to introduce -fsanitize flags?

Yes, OSS-fuzz sets all the necessary flags.

PR #230, opened 2021-01-23
Add reference to Debian and Ubuntu packages

roehling:
This PR adds a link to Debian und Ubuntu, which also ship the meshoptimizer library.


zeux:
Thanks!

PR #239, opened 2021-02-03
gltfpack: Implement support for KHR_materials_variants

### LINKED ISSUE #233: handling of KHR_materials_variants extension
Hi Guys,
Currently, gltfpack seems to be striping away everything related to KHR_materials_variants when such extension presents in the input file. Is this behavior intended?If it is intended, are you guys considering adding support for KHR_materials_variants extension in the future? If so, maybe I can help

zeux:
We now process all variant materials similarly to other materials and emit material mapping and variant data in the output.
Mesh merging is restricted to cases when the variant set is identical.

Fixes #233.

### DIFF
--- a/extern/cgltf.h
+++ b/extern/cgltf.h
@@ -492,6 +492,13 @@ typedef struct cgltf_material
 	cgltf_extension* extensions;
 } cgltf_material;
 
+typedef struct cgltf_material_mapping
+{
+	cgltf_size variant;
+	cgltf_material* material;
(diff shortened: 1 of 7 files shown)


zeux:
This isn't quite ready since when variant materials are used across different meshes, quantization bounds aren't going to be accurate. We need to effectively find material islands (when treating mesh-material relationship as a graph) to make sure any set of materials used in a single mesh has the exact same quantization bounds.

PR #282, opened 2021-05-14
Pass textures' samplers to optimized files

Jonek52:
Currently, meshoptimizer completely ignores information about textures' samplers stored in the glTF file. The attempt of this pull request is to fix that.

### DIFF
--- a/gltf/gltfpack.cpp
+++ b/gltf/gltfpack.cpp
@@ -413,8 +413,6 @@ static void process(cgltf_data* data, const char* input_path, const char* output
 		comma(json_samplers);
 		append(json_samplers, "{");
 		writeSampler(json_samplers, sampler);
-		if (settings.keep_extras)
-			writeExtras(json_samplers, extras, sampler.extras);
 		append(json_samplers, "}");
 	}
 
--- a/gltf/gltfpack.cpp
+++ b/gltf/gltfpack.cpp
@@ -406,7 +406,7 @@ static void process(cgltf_data* data, const char* input_path, const char* output
 	size_t mesh_offset = 0;
 	size_t material_offset = 0;
 
-	for (size_t i = 0u; i < data->samplers_count; ++i)
+	for (size_t i = 0; i < data->samplers_count; ++i)
 	{
 		const cgltf_sampler& sampler = data->samplers[i];
 
--- a/gltf/gltfpack.cpp
+++ b/gltf/gltfpack.cpp
@@ -376,6 +376,7 @@ static void process(cgltf_data* data, const char* input_path, const char* output
 	qt_dummy.bits = settings.tex_bits;
 
 	std::string json_images;
+	std::string json_samplers;
 	std::string json_textures;
 	std::string json_materials;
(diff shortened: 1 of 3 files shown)


gltf/gltfpack.cpp
```diff
@@ -405,6 +406,18 @@ static void process(cgltf_data* data, const char* input_path, const char* output
 	size_t mesh_offset = 0;
 	size_t material_offset = 0;
 
+	for (size_t i = 0u; i < data->samplers_count; ++i)
```
zeux, line 409:
```suggestion
	for (size_t i = 0; i < data->samplers_count; ++i)
```

gltf/gltfpack.cpp
```diff
@@ -405,6 +406,18 @@ static void process(cgltf_data* data, const char* input_path, const char* output
...
+	{
+		const cgltf_sampler& sampler = data->samplers[i];
+
+		comma(json_samplers);
+		append(json_samplers, "{");
+		writeSampler(json_samplers, sampler);
+		if (settings.keep_extras)
```
zeux, line 416: To keep extras for a given element, additional code is necessary during parsing. I would remove this for now, as gltfpack only adds support for extras with a solid use case and here it's not clear what the usecase is.

gltf/write.cpp
```diff
@@ -899,15 +911,18 @@ void writeTexture(std::string& json, const cgltf_texture& texture, cgltf_data* d
 {
 	if (texture.image)
 	{
+		append(json, "\"sampler\":");
```
zeux, line 914: sampler is optional and thus can be null: https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#reference-texture

gltf/write.cpp
```diff
@@ -899,15 +911,18 @@ void writeTexture(std::string& json, const cgltf_texture& texture, cgltf_data* d
...
+		append(json, size_t(texture.sampler - data->samplers));
+
 		if (settings.texture_ktx2)
 		{
-			append(json, "\"extensions\":{\"KHR_texture_basisu\":{\"source\":");
+			append(json, ",\"extensions\":{\"KHR_texture_basisu\":{\"source\":");
```
zeux, line 919: should use `comma` instead

zeux:
To help me understand the dominant use case, is it specifying clamp for wrapS/T to avoid bilinear filtering artifacts or something else?

gltf/gltfpack.cpp
Jonek52: Sure. Removed extras for now.

Jonek52:
> To help me understand the dominant use case, is it specifying clamp for wrapS/T to avoid bilinear filtering artifacts or something else?

Yes. It is good always good to specify clamping for wrapS/T explicite to avoid interpolation artifacts around the edges.

gltf/write.cpp
Jonek52: Added check for null sampler

zeux:
Thanks!

PR #283, opened 2021-05-17
fix checks in CMakeLists.txt

abma:
CMP0077 requires >= 3.13
CMP0091 + CMP0092 requires >= 3.15

we hit this bug here: https://github.com/spring/spring/pull/559/files#diff-48b508d8a5702095973f36f384be2a8b74834092bb7ed47c536f1af80015b20bR4

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,10 +1,10 @@
 cmake_minimum_required(VERSION 3.0)
 
-if (NOT CMAKE_VERSION VERSION_LESS 3.12)
+if (CMAKE_VERSION VERSION_GREATER_EQUAL 3.13)
     cmake_policy(SET CMP0077 NEW) # Enables override of options from parent CMakeLists.txt
 endif()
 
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -1,10 +1,10 @@
 cmake_minimum_required(VERSION 3.0)
 
-if (NOT CMAKE_VERSION VERSION_LESS 3.12)
+if (CMAKE_VERSION VERSION_GREATER_EQUAL 3.13)
```
mosra, line 3: Sorry for an unsolicited comment, but I thought this could help here. I'm usually doing just the following -- that way I don't need to look up what version the policy appeared in and so it's less error-prone:

```cmake
if(POLICY CMP0077)
    cmake_policy(SET CMP0077 NEW)
endif()
```

abma: oh, great, thats even better!

thanks!

zeux:
Thanks!

PR #285, opened 2021-05-19
KHR_lights_punctual: intensity light default to 1

scorpheus:
fix: when loading a gltf with "KHR_lights_punctual" extension with lights with no intensity (default 1), the gltf was saved with intensity 0 (instead of 1 or actually nothing because it's the default)


zeux:
This looks great, thanks! Would you mind also submitting this upstream to https://github.com/jkuhlmann/cgltf ?

PR #301, opened 2021-06-06
gltfpack: Internal Basis encoder

zeux:
This is an experimental change that embeds basisu as a library into
gltfpack. If this works well this could simplify the distribution story
for gltfpack since it would not rely on any seprate executables.

Drawbacks:

- Building this requires a basis_universal checkout. We could do this in
CI though.
- This is probably going to be painful to build for node.js builds.
- This will noticeably regress node.js compression performance due to
lack of SIMD or multithreading.

This change currently needs a few changes to upstream code to compile cleanly.

If anyone sees this and has thoughts on this they are very welcome. I'm pretty
tired of maintaining command line interfaces between gltfpack and toktx/basisu
so this might be a better alternative for me.


zeux:
This seems to more or less work now; the big caveat is that it's impossible to build node.js builds right now, since Basis uses std:: threading facilities that aren't working in WASI SDK at all...

Issue #313, opened 2021-07-08
Sampler data written incorrectly


dtrebilco:
https://github.com/zeux/meshoptimizer/blob/86740c27082cc75dd93623c003e9bc529564249e/gltf/write.cpp#L750

When reading in sampler data, all the fields are optional based on my reading of : 

https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#samplers 
"Default Filtering Implementation Note: When filtering options are defined, runtime must use them. Otherwise, it is free to adapt filtering to performance or quality goals."
and 
https://github.com/KhronosGroup/glTF/blob/master/specification/2.0/schema/sampler.schema.json

I have an example where the sampler filtering options are not specified on input, and are written as "0" on output that causes validation issues.
Input 
  "samplers": [
    {
      "minFilter": 9985
    },

Output
  "samplers": [
    {
      "magFilter": 0,
      "minFilter": 9985,
      "wrapS": 10497,
      "wrapT": 10497
    },

I think the sampler write code be updated to:

```c++
void writeSampler(std::string& json, const cgltf_sampler& sampler)
{
	if (sampler.mag_filter != 0)
	{
		comma(json);
		append(json, "\"magFilter\":");
		append(json, size_t(sampler.mag_filter));
	}
	if (sampler.min_filter != 0)
	{
		comma(json);
		append(json, "\"minFilter\":");
		append(json, size_t(sampler.min_filter));
	}
	if (sampler.wrap_s != 10497)
	{
		comma(json);
		append(json, "\"wrapS\":");
		append(json, size_t(sampler.wrap_s));
	}
	if (sampler.wrap_t != 10497)
	{
		comma(json);
		append(json, "\"wrapT\":");
		append(json, size_t(sampler.wrap_t));
	}
}
```

dtrebilco:
Pull request https://github.com/zeux/meshoptimizer/pull/314

PR #314, opened 2021-07-08
Fix for samplers being written out with default values and writing 0 …

dtrebilco:
…in filter fields if no filter value was read


dtrebilco:
This is for issue https://github.com/zeux/meshoptimizer/issues/313

zeux:
Thanks!

Issue #315, opened 2021-07-08
Meshes merged incorrectly 


dtrebilco:
I ran into an issue where multiple meshes had the node with flip face winding and some without, this caused some meshes to be baked inverted.

https://github.com/zeux/meshoptimizer/blob/86740c27082cc75dd93623c003e9bc529564249e/gltf/mesh.cpp#L252

In the mergeMeshInstances() method, this line:
```c++
	for (size_t i = 0; i < mesh.nodes.size(); ++i)
	{
		transformMesh(transformed, base, mesh.nodes[i]);
		mergeMeshes(mesh, transformed);
	}
```	
assumes that the transformed mesh is either completely overwritten by base mesh, or transformMesh() does not touch some fields.

This is not true, as in transformMesh() 
```c++
	if (det < 0 && mesh.type == cgltf_primitive_type_triangles)
	{
		// negative scale means we need to flip face winding
		for (size_t i = 0; i < target.indices.size(); i += 3)
			std::swap(target.indices[i + 0], target.indices[i + 1]);
	}
```	
directly manipulates the index buffer conditionally. 

What needs to happen is the index buffer to be re-copied every iteration. 
```
target.indices = mesh.indices;
```
This could happen in transformMesh() or in the outer loop in mergeMeshInstances()

PR #316, opened 2021-07-08
Fix merging multiple meshes that have different flip face winding

dtrebilco:
Fix for issue https://github.com/zeux/meshoptimizer/issues/315


zeux:
Thanks!

PR #318, opened 2021-07-22
The PNG image type should be at offset 25 - not 27

dtrebilco:
https://en.wikipedia.org/wiki/Portable_Network_Graphics#File_header
8 signature
8 IHDR start
4(width) + 4(height) + 1 bit depth
26th byte is color type (index 25)


zeux:
Yup :( there was an off-by-2 error for other offsets before in this code as well, not sure how exactly that happened.

PR #319, opened 2021-07-22
Optimize output types

dtrebilco:
Do not write out several options if at the default values.
Texture view transforms were a bit difficult as several places check for the extension being used - I broke "const" in updating the values during the material analyse phase as I could not easily see another obvious place to change this.

Sorry this also includes the PNG change, I could not figure out how to cherry pick in github easily.

### DIFF
--- a/gltf/gltfpack.cpp
+++ b/gltf/gltfpack.cpp
@@ -529,7 +529,7 @@ static void process(cgltf_data* data, const char* input_path, const char* output
 			append(json_meshes, "{\"attributes\":{");
 			writeMeshAttributes(json_meshes, views, json_accessors, accr_offset, prim, 0, qp, qt, settings);
 			append(json_meshes, "}");
-			if (prim.type != 4)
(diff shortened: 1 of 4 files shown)


zeux:
How valuable are different components of this change? Iirc some of these like byteOffsets were intentional for simplicity and data uniformity - e.g. having a single accessor omit byte offset isn't very beneficial wrt overall file size. So curious as to what the motivation here is.

dtrebilco:
Most of these were just found while I was diffing pre-post changes to find other bugs.
The only one I really care about is the texture_view change. 

An exporter we use is generating these texture views on every material texture slot. This causes downstream code to have to do custom logic on every material that I would like to avoid. 

If you reject this change it is OK - I can handle it elsewhere.

zeux:
Coming back to this, I've updated the PR a bit to remove the parts that I feel like aren't very important. The rest looks good to merge; patching the views during analysis looks a little bit odd, but I'll refactor it post-merge as it's too difficult to do in the web interface :)

PR #322, opened 2021-07-28
js: Implement a standalone encoder library

zeux:
This module can be used to compress meshes, most commonly to conform to EXT_meshopt_compression (https://github.com/KhronosGroup/glTF/blob/master/extensions/2.0/Vendor/EXT_meshopt_compression/README.md).

This change currently lacks documentation, this will be added in the future.

The recommended vanilla integration for geometry compression looks like this:

1. Reorder mesh index/vertex buffers, using `reorderMesh(indices, true, true)` (assuming triangle meshes) which returns a remap table that can be used to create new vertex buffers - remap table maps *old* indices to *new* so the remap loop for vertex data looks smth like `newvertices[remap[i]] = oldvertices[i];`. This improves compression in the following step.
2. Encode all vertex streams using `encodeVertexBuffer`; encode all index buffers using `encodeIndexBuffer` (note that vertex encoding works with Uint8Array but index encoding works with Uint16/Uint32Array, I can change that if necessary) if they contain triangles, and `encodeIndexSequence` otherwise
3. Serialize glTF extension metadata, using "ATTRIBUTES" mode for vertex data, "TRIANGLES" for IndexBuffer encoding and "INDICES" for IndexSequence

This works equally well with interleaved and non-interleaved vertex data; quantization should be applied before encoding.

When encoding other glTF streams for animation etc., encodeVertexBuffer can be used in the same way on the input stream (without any reordering).

This provides basic encoding for everything that should be roughly equivalent to `gltfpack -c`. When more compression is required, encode filters come into play - using these should make it possible to match `gltfpack -cc` in terms of transmission size. The encode filters can be used to preprocess floating point data and generate special quantized streams for three types of data: unit vectors (normals/tangents), quaternions (animation or instance rotations) and general floating point data (translation and scale vectors, and unquantized positions/texture coordinates if necessary).

To apply the filters, `encodeFilter*` should be used to produce a Uint8Array from input Float32Array. The resulting filtered buffer can be encoded using `encodeVertexBuffer` - if the glTF extension metadata then specifies this filter, the decoder will reverse the filtering operation and produce optimal quantized data for unit vectors or rotations, or floating-point data for other types. The type of filter should be saved to "filter" attribute in the glTF extension blob so that the decoder knows which filter to reverse.

Note that filters are lossy - which is why the number of bits to use per component is a required argument; I recommend 8 bits for normal/tangent data (10 bit for high quality normals which requires 8-byte output), 12 bits for quaternions for rotations, and ~15 bits for general floating point storage. Examples for using filters are in the .test.js file in this PR and they will be added to the documentation. When filters are not used, the encoding is lossless - mesh reordering only changes the order of rendered triangles, the compression is lossless by itself. Filters are used to reduce entropy and prepare data in the optimal fashion for the subsequent compression to take place, which is also why they aren't really necessary for data that can already be optimally quantized using integers.

There are some restrictions on the input stride/bits that are currently checked in C++ implementation but aren't checked in the JS wrappers which may make some errors more difficult to debug. I'm going to add validation and try to convert the text above + information on how to use the decoder into a README file, any feedback is welcome otherwise.

cc @donmccurdy


donmccurdy:
Looking forward to trying this out! Pretty shocked this is only 6.5kb minzipped. I expect I'll be able to work on it next week, but do have a couple questions in the meantime:

> vertex encoding works with Uint8Array but index encoding works with Uint16/Uint32Array, I can change that if necessary

In gltfpack `-vp` defaults to 14 bits, I am a bit nervous about allowing only 8 bits when filters are not used... or is the idea that a position attribute using Uint16Array can just be re-wrapped in a Uint8Array with `size=6`?

My [quantization implementation](https://github.com/donmccurdy/glTF-Transform/blob/b4c39a3aabaf39e537b8e6c06d2044111552db2a/packages/functions/src/quantize.ts#L293-L301) currently uses Int8Array and Int16Array (rather than the unsigned alternatives) for several attribute types, but there's no particular reason for that and I could switch to Uint.

> This works equally well with interleaved and non-interleaved vertex data

I'm not sure what arguments to `encodeVertexBuffer` would be used for interleaved vertex data, not seeing a `stride`, have I missed something there?

donmccurdy:
Just curious – does gltfpack ever write sparse accessors, e.g. for morph attributes? I'm assuming we'd want to keep the attribute sparse if doing so was already advantageous without compression, remap the indices to the new order, and call `encodeVertexBuffer` with appropriate count?

zeux:
>  I am a bit nervous about allowing only 8 bits when filters are not used... or is the idea that a position attribute using Uint16Array can just be re-wrapped in a Uint8Array with size=6?

Yeah that's exactly it. Would it be more idiomatic to use ArrayBuffer instead of Uint8Array here? I use Uint8Array right now as "byte array" essentially.

> I'm not sure what arguments to encodeVertexBuffer would be used for interleaved vertex data, not seeing a stride, have I missed something there?

You'd just pass a larger `size`, so `size` and `stride` here have the same meaning.

donmccurdy:
> Would it be more idiomatic to use ArrayBuffer instead of Uint8Array here? I use Uint8Array right now as "byte array" essentially.

Got it – yes, I think Uint8Array is the right choice as long as (1) it's clear in documentation that it's being used as a byte array, and (2) there's no requirement that the underlying ArrayBuffer be the same size as the array.

> You'd just pass a larger size, so size and stride here have the same meaning...

Oops, I had misunderstood this. So `encodeVertexBuffer` is not being called for each vertex attribute necessarily. For an interleaved primitive with three vec3 uint16 attributes, we'd pass a Uint8Array containing the already-interleaved data, and a size of `3 * 3 * 2` (plus padding, where required)?

zeux:
> So encodeVertexBuffer is not being called for each vertex attribute necessarily.

Yeah - correct - you call it for the bufferView contents, and pass the size of a single "element", which for interleaved data would be "all attributes of a given vertex". I'm going to add `encodeGltfBuffer` helper as well to be more symmetrical with decoder and to make it a bit cleaner.

zeux:
I've adjusted the interface for all functions to hopefully be final (changed index encoding to also take in a byte view and added assertions for preconditions to simplify usage). Will try to add documentation tomorrow and commit this.

zeux:
I've done my best to document both decoders and encoders from the API perspective. I'm going to merge this, but feel free to raise issues in the documentation or the actual implementation since glTF-Transform is going to be the first non-C++ implementation that uses these.

js/meshopt_encoder.js
```diff
@@ -0,0 +1,182 @@
...
+		target.set(heap.subarray(tp, tp + count * stride));
+		sbrk(tp - sbrk(0));
+		return target;
+	}
+
+	return {
+		ready: promise,
+		supported: true,
+		reorderMesh: function(indices, triangles, optsize) {
+			var optf = triangles ? (optsize ? instance.exports.meshopt_optimizeVertexCacheStrip : instance.exports.meshopt_optimizeVertexCache) : undefined;
```
donmccurdy, line 123: Is this intended? Setting `triangles=true, optsize=true` works as expected, but `triangles=true, optsize=false` results in a broken triangle mesh.

zeux: One thing I recommend with this entire integration in general is to check each component independently:

- You can encode the mesh without using filters and without reordering
- You can reorder the mesh without using encoding
- After that you can separately apply filters if need be :)

Are you seeing broken meshes with just reorder being called? If you can share your code I'll be happy to debug it - there are unit tests for this but they may be incomplete, or the calling code may be subtly wrong.

donmccurdy: Ah sorry, found the bug (my mistake) halfway through setting up the fiddle. Yes, I'm starting with reordering alone, since it could be useful as a separate step in glTF-Transform.

PR #325, opened 2021-08-08
js: Update meshoptimizer NPM package

zeux:
This adds missing .ts files and adds index.* so that this package is a bit easier to use.


zeux:
@donmccurdy I've published this as 0.16.1

donmccurdy:
Thanks! Switched over in https://github.com/donmccurdy/glTF-Transform/pull/322; both CommonJS and ES Module versions are working for me, as well as type definitions.

PR #331, opened 2021-08-25
js: Add reference decoder by Jasper

zeux:
This is the reference decoder implementation.
It follows the decoder interface and should be a drop-in replacement for the actual decoder from meshopt_decoder.js.

It is provided for educational value and is not recommended for use in production because it's not performance-optimized.


abhinay17644:
XKjNI2sE4grK_1080x1920

PR #399, opened 2022-01-18
gltfpack: Sort bone influences by weight, not by index

zeux:
This slightly regresses the compression ratio but the delta is not very significant -
it seems to be <~0.1% on most meshes and goes up to ~0.3% on some.

We can probably recoup some of the losses by sorting the skin joints in the future
by making the neighboring joints more spatially proximal wrt rigging.

With this change renderers that discard lower weights and renormalize the rest will
not have to sort the weights for glTF files packed by gltfpack in the future, even
though there's no way to indicate that in the file for now.

### DIFF
--- a/gltf/mesh.cpp
+++ b/gltf/mesh.cpp
@@ -593,14 +593,6 @@ struct BoneInfluence
 	float w;
 };
 
-struct BoneInfluenceIndexPredicate
-{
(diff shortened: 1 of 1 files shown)


zeux:
This partially reverts https://github.com/zeux/meshoptimizer/commit/b54af103ac8f509562a2a576b56514ad10341518

PR #430, opened 2022-06-09
CMake: Updated install script, added COMPONENT.

DragonJoker:
This may be needed when creating setup packages.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -117,20 +117,21 @@ endif()
 include(GNUInstallDirs)
 
 install(TARGETS ${TARGETS} EXPORT meshoptimizerTargets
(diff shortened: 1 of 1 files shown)


zeux:
Can you clarify why this is needed? I'm used to occasionally having to specify different components for different parts of a large library but it's not obvious to me why it's worthwhile to assign the same component to all install targets.

DragonJoker:
When you use CPack, all install scripts without a component will go to "Unspecified", instead of (in that case) "meshoptimizer", that's why I've introduced this thing.

zeux:
Got it, thanks. It will probably make sense to use `gltfpack` component for `gltfpack` target, but this change by itself should be fine for now...

PR #443, opened 2022-07-24
gltfpack: Implement support for normalized attributes

### LINKED ISSUE #422: Add gltfpack quantization mode that creates normalized position accessors
At the moment gltfpack, when quantizing meshes, creates non-normalized vertex position accessors and counter-acts the large vertex position values with small node scale values. Some engines can run into precision problems with small scale values like this. Unity is an example and one where this [cannot be easily fixed](https://issuetracker.unity3d.com/issues/objects-get-rendered-black-slash-dark-color-when-their-scale-is-ridiculously-small).

A solution would be to create accessors with normalized set to true, and thus not having to have very small values in a node’s transform.

@zeux in a slack discussion (on the glTF slack) you mentioned the status quo is due to three.js having trouble with normalized accessors. Ideally three.js fixed this issue by now, otherwise it would be great having a switch/parameter for gltfpack to force it to create normalized accessors instead, so Unity users can benefit of the great gltfpack as well.

Thanks for considering!

P.S.: Created this issue, so I can link to it from affected issues.

zeux:
This change adds -vpn flag that enables the use of normalized position
accessors instead of integer (unnormalized) ones, and switches texture
coordinates to normalized attributes unconditionally.

Currently gltfpack quantizes positions and texture coordinates using
integers; this results in very small scaling values that are required to
transform the values back into 0..1 range.

While this is generally not a problem, there are some contexts in which
the scale values are too small and may be truncated to 0, eg if the shader
is using medium precision. Additionally, for native GPU backends like
WebGPU the shader may need to match the logical type of the vertex
attribute with the shader declaration - for unnormalized integers to
work, one must use i32/u32 types, which makes meshes that use
quantization incompatible with shaders that expect non-quantized inputs.

Using normalized data fixes this. For texture coordinates, we should be
able to do that unconditionally. For positions, this can have a downside
in that it requires an application of an extra normalization factor
whenever you work with the data on the CPU side, which can break
frameworks like Three.JS that can read data directly out of Int16Array
otherwise. Additionally, this mode reduces precision for morph deltas by
one bit since we need to equalize the deltas between snorm and unorm.

While the morph target issue could be solved by using normalized signed
attributes for positions, they may decode with precision issues on Web and
as such we avoid them out of caution. We still have to use them for morph
targets though, so `-vpn` may have further unintended consequences for
Web platform when morph targets are used.

A small downside of this change is a small precision loss during
reconstruction - previously we did the reconstruction in one step, thus
losing ~0.5ulp of precision on that, and now reconstruction is two-step:
GPU divides the integer by 65535, and then we rescale back, which may
result in artifacts - however, in practice the extra 0.5ulp is always
insignificantly small to be observable, and even for positions it should never
result in gaps that weren't there previously.

Fixes #422.

### DIFF
--- a/gltf/gltfpack.cpp
+++ b/gltf/gltfpack.cpp
@@ -657,7 +657,7 @@ static void process(cgltf_data* data, const char* input_path, const char* output
 				assert(ni.keep);
 				ni.meshes.push_back(node_offset);
 
-				writeMeshNode(json_nodes, mesh_offset, mesh.nodes[j], mesh.skin, data, settings.quantize ? &qp : NULL);
+				writeMeshNode(json_nodes, mesh_offset, mesh.nodes[j], mesh.skin, data, settings.quantize ? &qp : NULL, settings);
 
 				node_offset++;
 			}
@@ -681,7 +681,7 @@ static void process(cgltf_data* data, const char* input_path, const char* output
 			comma(json_roots[mesh.scene]);
 			append(json_roots[mesh.scene], node_offset);
 
-			writeMeshNode(json_nodes, mesh_offset, NULL, mesh.skin, data, settings.quantize ? &qp : NULL);
+			writeMeshNode(json_nodes, mesh_offset, NULL, mesh.skin, data, settings.quantize ? &qp : NULL, settings);
 
 			node_offset++;
 		}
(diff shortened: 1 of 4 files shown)


atteneder:
Awesome! Thanks a lot!

PR #445, opened 2022-07-24
gltfpack: Optimize .obj parsing for memory consumption

zeux:
With this change we now deduplicate face vertices before creating the
vertex data, and don't create streams that we know don't exist.

This results in memory savings for loading large .obj files, although
this also regresses some:

Before:
./gltfpack -i /home/zeux/Downloads/rungholt.obj -o rh.glb 3.84 sec 924576 KB
./gltfpack -i /home/zeux/Downloads/hairball.obj -o hb.glb 1.36 sec 600440 KB
./gltfpack -i /home/zeux/Downloads/splash.obj -o sp.glb 15.12 sec 4249600 KB

After:
./gltfpack -i /home/zeux/Downloads/rungholt.obj -o rh.glb 3.91 sec 1100132 KB
./gltfpack -i /home/zeux/Downloads/hairball.obj -o hb.glb 1.26 sec 275084 KB
./gltfpack -i /home/zeux/Downloads/splash.obj -o sp.glb 12.06 sec 2766064 KB

It's not clear where the regression on rungholt comes from, although the
peak memory consumption there seems to be driven by mesh merging
instead.

### DIFF
--- a/gltf/parseobj.cpp
+++ b/gltf/parseobj.cpp
@@ -2,8 +2,7 @@
 #include "gltfpack.h"
 
 #include "../extern/fast_obj.h"
-
-#include <algorithm>
+#include "../src/meshoptimizer.h"
 
 #include <stdlib.h>
 #include <string.h>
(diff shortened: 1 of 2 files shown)


zeux:
Looking closer at rungholt, I'm not overly concerned with this as the gain doesn't show up when measuring allocation usage, so it's highly allocator specific, might be related to fragmentation. Also that was on the file without .mtl loaded, with .mtl the baseline memory consumption is higher and the relative increase is lower. Overall this change seems important and beneficial, we can investigate other memory reductions separately.

zeux:
I've double checked and the increase is indeed specific to the system allocator used; I don't think it's fragmentation-related and instead must be related to the policy for freeing pages to the OS. Using mimalloc as the allocator actually reverses the results, with the peak RSS being *lower* after this change.

zeux:
A few other large files show improvement as expected, eg powerplant.obj:

Before: 6.56 sec 2575132 KB
After: 5.49 sec 1250440 KB

PR #446, opened 2022-07-25
js: Add Wasm simplifier

### LINKED ISSUE #427: Mesh Simplify available as WASM
Is this available as a WASM module at this time?  How much work is it to convert to a WARM module like the other components that are already realized as WASM?

(This is an amazing tool overall, just trying to make it even more accessible.)

zeux:
This change exposes simplifier through a JS module, similarly to
encoder/decoder. For now we only expose the main simplifier and don't
expose sloppy / point cloud versions.

demo/simplify.html has a simple integration demo into Three.JS although of
course this isn't a very efficient method of updating the buffers (eg we don't
compact the mesh using compactMesh function so the entire original mesh
is always resident in memory - that said, this is a reasonable way to generate
LOD for meshes after loading is done).

Fixes #427


zeux:
The 16/32-bit index conversion should probably be done automatically in simplify() function...

PR #454, opened 2022-08-07
js: Implement support for WebWorker-driven decoding

zeux:
This change implements support for using WebWorkers during mesh
decoding; MeshoptDecoder module now has a function, useWorkers, that -
when called - switches behavior of the new function
decodeGltfBufferAsync to use web workers.

decodeGltfBufferAsync can also work without web workers, and it doesn't
require the use of ready promise, which simplifies support in GLTF
loaders.

### DIFF
--- a/js/meshopt_decoder.js
+++ b/js/meshopt_decoder.js
@@ -105,12 +105,7 @@ var MeshoptDecoder = (function() {
 			var data = event.data;
 
 			worker.pending -= data.count;
-
-			if (data.error) {
-				worker.requests[data.id].reject(data.error);
-			} else {
-				worker.requests[data.id].resolve(data.target);
-			}
+			worker.requests[data.id][data.action](data.value);
 
 			delete worker.requests[data.id];
 		};
@@ -122,8 +117,8 @@ var MeshoptDecoder = (function() {
 		var source =
 			"var instance; var promise = WebAssembly.instantiate(new Uint8Array([" + new Uint8Array(unpack(wasm)) + "]), {})" +
 			".then(function(result) { instance = result.instance; instance.exports.__wasm_call_ctors(); });" +
-			"self.onmessage = processWorker;" +
-			decode.toString() + processWorker.toString();
+			"self.onmessage = workerProcess;" +
+			decode.toString() + workerProcess.toString();
 
 		var blob = new Blob([source], {type: 'text/javascript'});
 		var url = URL.createObjectURL(blob);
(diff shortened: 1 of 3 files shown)


zeux:
This requires a change in GLTFLoader to use the new function, something like this is compatible with the old versions of the module:

```
				if ( decoder.decodeGltfBufferAsync ) {

					return decoder.decodeGltfBufferAsync( count, stride, source, extensionDef.mode, extensionDef.filter ).then( function ( res ) {

						return res.buffer;

					} );

				} else {

					return decoder.ready.then( function () {

						var result = new ArrayBuffer( count * stride );
						decoder.decodeGltfBuffer( new Uint8Array( result ), count, stride, source, extensionDef.mode, extensionDef.filter );
						return result;

					} );

				}
```

This change by itself increases the size of the decoder.js by ~10% post-gzip (6 KB -> 6.6 KB). I need to look into whether it can be simplified to reduce the size impact, even though 600 bytes is not the end of the world.

zeux:
cc @Ben-Mack @wallabyway

zeux:
It doesn't look like the size impact could be reduced realistically; it's consistently 10% (+600 bytes) after gzip. When using minification, the old minified .js is 5500 bytes post-gzip, the new one is 6000 bytes, give or take.

For now I'd say this is a modest increase that doesn't justify a separate module for the implementation; if we need to, we can always strip dead code and remove WebWorkers support, or maybe tree sitter does that automatically in production deployments.

wallabyway:
I tried using this, but got some weird results...

I modified glTFloader...

GLTFLoader.js
```
return Promise.all( [ buffer, decoder.ready ] ).then( function ( res ) {

				const byteOffset = extensionDef.byteOffset || 0;
				const byteLength = extensionDef.byteLength || 0;

				const count = extensionDef.count;
				const stride = extensionDef.byteStride;

				const result = new ArrayBuffer( count * stride );
				const source = new Uint8Array( res[ 0 ], byteOffset, byteLength );

				//decoder.decodeGltfBuffer( new Uint8Array( result ), count, stride, source, extensionDef.mode, extensionDef.filter );
				//return result;
				// async web worker decoder
				const resultAsync = await decoder.decodeGltfBufferAsync( count, stride, source, extensionDef.mode, extensionDef.filter );
				return resultAsync;
} );
```

 and used this experimental `meshopt_decoder.module.js web-worker version, but the promise, returns buffers, that came back as scaled to ~1/10th their geometry size.  Kinda weird.

Probably something I did.  Anyway, I tried it.  I noticed a small difference in main thread UI stalls, but probably not enough to warrant using web-workers.  I'll return to this at some point and try again.

zeux:
Note that you should be able to use latest three.js (r144+ to be precise) and there's no need to change GLTFLoader.js, but you do need to call MeshoptDecoder.useWorkers in either case to activate that. My experience has been that on Linux this eliminates most geometry related stalls, but on Windows there's some unavoidable cost in WebGL implementation in Chrome for uploading buffer data - this will probably improve with WebGPU in the future.

PR #455, opened 2022-08-07
vertexcodec: Use a more efficient lowering for neonMoveMask

zeux:
Instead of using horizontal adds or more expensive paired adds on 32-bit ARM,
we can use the same trick we use for Wasm which assumes that all input
components are 32-bit and uses a specially crafted constant to turn the result
into a bitmask.

This results in ~2% faster decoding on Apple M2; the performance will need to
be validated on other ARM CPUs.

### DIFF
--- a/src/vertexcodec.cpp
+++ b/src/vertexcodec.cpp
@@ -607,7 +607,7 @@ static void neonMoveMask(uint8x16_t mask, unsigned char& mask0, unsigned char& m
 	// magic constant found using z3 SMT assuming mask has 8 groups of 0xff or 0x00
 	const uint64_t magic = 0x000103070f1f3f80ull;
 
-	uint64x2_t mask2 = vreinterpretq_u8_u64(mask);
+	uint64x2_t mask2 = vreinterpretq_u64_u8(mask);
 
 	mask0 = uint8_t((vgetq_lane_u64(mask2, 0) * magic) >> 56);
 	mask1 = uint8_t((vgetq_lane_u64(mask2, 1) * magic) >> 56);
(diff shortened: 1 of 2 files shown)


zeux:
This is ~3% faster on Amazon Graviton 2, so this seems to be an overall good idea on AArch64. I need to test this on ARMv7 since it doesn't support 64-bit integers and as such the codegen is not very inspiring.

zeux:
For AArch64, llvm-mca predicts that the scalar sequence is better than horizontal adds (~26 cycles/op for existing code, ~18 cycles/op for scalar code, CPU is Cortex-A53)

For ARMv7, llvm-mca predicts that the scalar sequence is worse than paired adds (~23 cycles/op for existing code, ~29 cycles/op for scalar code, CPU is Cortex-A9)

Since 64-bit results match the observed performance on two different CPUs, I'm inclined to believe the 32-bit results either. This instruction isn't always critical to processing time so maybe the loss (23->29 cycles) is going to be negligible but it's probably better to keep the existing sequence for 32-bit ARM specifically... I'll need to get an actual hardware with a 32-bit ARM CPU to verify perf impact.

zeux:
This is also likely massively better for MSVC, which currently uses the paired-add path since horizontal adds aren't exposed; llvm-mca says that on Cortex-A53, current MSVC code is 47 cycles/op, and new scalar code is ~14 cycles/op.

zeux:
Running this on ARM Cortex-A53 in 32-bit mode shows no performance change between vectorized variant with pairwise adds and scalar variant, so I'm going to merge this as is as it benefits 64-bit systems without regressing 32-bit ones.

Issue #433, opened 2022-07-09
gltfpack produces broken result for skinned parts of mesh


hybridherbst:
1. Download https://github.com/prefrontalcortex/glTF-Sample-Models/raw/carbon-frame-bike/2.0/CarbonFrameBike/glTF-Binary/CarbonFrameBike.glb
2. See reference images at https://github.com/prefrontalcortex/glTF-Sample-Models/tree/carbon-frame-bike/2.0/CarbonFrameBike
3. Run `gltfpack -i CarbonFrameBike.glb -o CarbonFrameBike.packed.glb`
4. Open the result in e.g. [Babylon Sandbox](https://sandbox.babylonjs.com/) or Gestaltor
5. Note that the wires don't survive packing well:
  ![image](https://user-images.githubusercontent.com/2693840/178118767-575834d6-8445-408a-8025-00618e8fcf6c.png)

Reference: 
![image](https://user-images.githubusercontent.com/2693840/178118964-dd1a972b-553c-49a7-82f5-e639de0e53b7.png)

Using `-noq` works around the problem, but I thought it would be worth still raising an issue since the result looks relatively broken with default parameters.

zeux:
fwiw this looks as if the default number of bits (14) gltfpack uses for quantization isn't sufficient for this model. In certain cases this can be avoided by recentering the mesh, which gltfpack doesn't do automatically, but often it's just something that needs to be raised via `-vp` command line flag.

I'll keep this open for a bit because I need to check if gltfpack can improve this automatically by changing the mesh parameters, but using something like `-vp 16` is likely to improve this noticeably.

zeux:
Right, so:

1. This can be improved via `-vp 16`
2. This happens because the wires in question have a very large internal scale value (of 100 - this looks like an improper coordinate system conversion somewhere; probably the inverse scale of 1/100 was baked into the vertex data). Since quantization is done in local space, but using a single global quantization grid (relaxing this is difficult because eg several meshes can reuse skin objects and the quantization scale needs to be corrected via skinning transforms), the vertices end up being quantized with suboptimal values.

I think you don't see the issue in glTF-Transform because it's not quantizing the values, either because it just doesn't support quantization on skinned meshes, or because it requires an explicit command line instruction to enable that.

Either way, I don't think this can realistically be fixed in quantization code, and propagating the scale value "into" mesh vertices would result in different bone transforms which may be important for other applications.

zeux:
[...]

PR #462, opened 2022-08-17
gltfpack: Experimental support for floating point position quantization

zeux:
Currently gltfpack provides all or nothing quantization: either all
attributes get quantized, or none do.

When quantization is used, we use the most size and memory efficient
formats, which is great, but it can lead to integration issues for
applications - since dequantization uses an extra transform, this
results in changes that are needed for some applications to take that
transform into account. Notably, these changes mostly revolve around
positions (and rarely around texture coordinates wrt texture
replacement, but positions are definitely the main source of issues).

Due to the flexibility of meshopt codecs however, there are other
options available that don't change the coordinate space of mesh
positions, but instead just change the positions a little bit to encode
more efficiently.

This change adds a new mode, -vpf, which uses floating point
quantization to encode position values:

- When using regular compression (-c), we quantize each coordinate
  individually
- When using extra compression (-cc), we use exponential filter when
  serializing the position stream

It is expected that either mode is less efficient than integer
quantization, but -cc can be closer to integer quantization and on
meshes with many attributes the extra size could be an acceptable
tradeoff for better compatibility. Of course this also results in geometry
taking a little bit more space after being loaded, since we use 12 bytes
per position instead of 8 bytes.

- [x] Fix min/max bounds (these need to be precise to validate correctly)
- [x] Gather size statistics for various models
- [x] Validate quantization bits behavior for models to ensure precision is reasonably close between different modes
- [x] Figure out how to select between float quantization vs separate vs shared exponent

### DIFF
--- a/gltf/gltfpack.cpp
+++ b/gltf/gltfpack.cpp
@@ -507,7 +507,7 @@ static void process(cgltf_data* data, const char* input_path, const char* output
 
 		comma(json_materials);
 		append(json_materials, "{");
-		writeMaterial(json_materials, data, material, settings.quantize ? &qp : NULL, settings.quantize ? &qt_materials[i] : NULL);
+		writeMaterial(json_materials, data, material, settings.quantize && !settings.pos_float ? &qp : NULL, settings.quantize ? &qt_materials[i] : NULL);
 		if (settings.keep_extras)
 			writeExtras(json_materials, extras, material.extras);
 		append(json_materials, "}");
@@ -657,7 +657,7 @@ static void process(cgltf_data* data, const char* input_path, const char* output
 				assert(ni.keep);
 				ni.meshes.push_back(node_offset);
 
-				writeMeshNode(json_nodes, mesh_offset, mesh.nodes[j], mesh.skin, data, settings.quantize ? &qp : NULL);
+				writeMeshNode(json_nodes, mesh_offset, mesh.nodes[j], mesh.skin, data, settings.quantize && !settings.pos_float ? &qp : NULL);
 
 				node_offset++;
 			}
@@ -681,7 +681,7 @@ static void process(cgltf_data* data, const char* input_path, const char* output
 			comma(json_roots[mesh.scene]);
 			append(json_roots[mesh.scene], node_offset);
 
-			writeMeshNode(json_nodes, mesh_offset, NULL, mesh.skin, data, settings.quantize ? &qp : NULL);
+			writeMeshNode(json_nodes, mesh_offset, NULL, mesh.skin, data, settings.quantize && !settings.pos_float ? &qp : NULL);
 
 			node_offset++;
 		}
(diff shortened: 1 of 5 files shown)


zeux:
TODO:

- Gather size statistics for various models
- Validate quantization bits behavior for models to ensure precision is reasonably close between different modes
- Check if exponential encoding (but one coordinate at a time) is better for `-c`

zeux:
Using very aggressive 8 bits for position (default is 14) on SciFiHelmet:

-vp 8 -cc (current), ~10.1 bits per position
image

-vp 8 -vpf (new, float quantization), ~36 bits per position
image

-vp 8 -vpf -c (new, separate exponent encoding), ~21.7 bits per position
image

-vp 8 -vpf -cc (new, shared exponent encoding), ~14.3 bits per position
image

It looks like shared exponent might be too aggressive to pursue, since it introduces a significant amount of error due to coupled components - this gets worse because gltfpack likes to transform meshes to world space, which doesn't matter for quantization but does matter for this new mode (and might be something to disable conditionally...)

zeux:
BrainStem.gltf - again, with 8-bit quantization (very aggressive):

-vp 8 -cc (quantized integer): 10 bits
image

-vp 8 -vpf (quantized float): 38 bits
image

-vp 8 -vpf -c (separate exponent): 25 bits
image

-vp 8 -vpf -cc (shared exponent): 14 bits
image

Here shared exponent seems to be doing alright, probably because the mesh is more local / centered which is normal for rigged meshes.

zeux:
Because quantized compressed positions can be misleading wrt the resulting size, here's just the size stats for the two models above with default bits (14):

SciFiHelmet:
-cc (integer quantization): 32.7 bits
-vpf (float quantization): 47.9 bits
-vpf -c (separate exponent): 44.3 bits
-vpf -cc (shared exponent): 39.6 bits

BrainStem:
-cc (integer quantization): 29.8 bits
-vpf (float quantization): 44.8 bits
-vpf -c (separate exponent): 43.4 bits
-vpf -cc (shared exponent): 32.8 bits

So that I don't get confused in the future: all numbers for float quantization do include vertex codec compression on top even if the current version of the code doesn't allow that configuration.

Here's what the data looks like for the entire glTF buffer (sans textures) after zstd:

SciFiHelmet:
-cc (integer quantization): 562 KB
-vpf (float quantization): 670 KB
-vpf -c (separate exponent): 650 KB
-vpf -cc (shared exponent): 608 KB

BrainStem:
-cc (integer quantization): 254 KB
-vpf (float quantization): 310 KB
-vpf -c (separate exponent): 305 KB
-vpf -cc (shared exponent): 265 KB

zeux:
CarbonBike model from #433 (the new mode isn't sensible to local transforms to the same extent and as such the model is preserved well), before gzip/zstd:

no quantization: vertex 3208635 bytes, index 126219 bytes
integer quantization: vertex 848142 bytes, index 126219 bytes
float quantization: vertex 1186861 bytes, index 126219 bytes
separate exponent: vertex 1166609 bytes, index 126219 bytes
shared exponent: vertex 1049273 bytes, index 126219 bytes

Overall this feels like a great future replacement for `-noq` but I'll need to figure out how to think about shared/separate exponent given the precision concerns vs size tradeoff.

zeux:
One other alternative is to share the exponent between all components of a single coordinate in a vertex stream. This is close in spirit to uniform quantization, as it essentially selects a power-of-two bucket and thus should be close to the error of the integer quantization, but the exponent doesn't take space in the stream as it compresses perfectly.

8 bits + integer quantization: 10.1 bits:

image

8 bits + float quantization: 36 bits:

image

8 bits + separate exponent: 20.4 bits:

image


8 bits + shared exponent: 14.3 bits:

image

8 bits + component exponent: 14.2 bits:

image

PR #472, opened 2022-09-02
Add vcpkg installation instructions

LilyWangLL:
Meshoptimizer is available as a port in [vcpkg](https://github.com/microsoft/vcpkg), a C++ library manager that simplifies installation for meshoptimizer and other project dependencies. Documenting the install process here will help users get started by providing a single set of commands to build meshoptimizer, ready to be included in their projects.

We also test whether our library ports build in various configurations (dynamic, static) on various platforms (OSX, Linux, Windows: x86, x64, UWP, ARM) to keep a wide coverage for users.

I'm a maintainer for vcpkg, and [here](https://github.com/microsoft/vcpkg/blob/master/ports/meshoptimizer/portfile.cmake) is what the port script looks like. We try to keep the library maintained as close as possible to the original library.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,17 +39,7 @@ The source files are organized in such a way that you don't need to change your 
 
 ## Installing from vcpkg
 
-The meshoptimizer port in vcpkg is kept up to date by Microsoft team members and community contributors. The url of vcpkg is: https://github.com/Microsoft/vcpkg . You can download and install meshoptimizer using the vcpkg dependency manager:
-
-```shell
-git clone https://github.com/Microsoft/vcpkg.git
-cd vcpkg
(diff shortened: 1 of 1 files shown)


zeux:
Thanks! Are the vcpkg installation instruction available via some sort of permalink in vcpkg repository / documentation site? I'd rather not cite them, both to save space as I'd assume people who would want to use vcpkg with meshoptimizer already have it installed, but also to allow for nuance - the instructions in this PR won't work on Windows.

LilyWangLL:
Thanks for your reply. Vcpkg has the documentation of instruction: https://github.com/microsoft/vcpkg#getting-started, I simplified this introduction, and attached the installation instructions url on vcpkg. Could you approve the new change about this?

zeux:
Thanks!

PR #504, opened 2022-10-27
gltfpack: Calculate AABB of morph targets correctly

### LINKED ISSUE #503: AABB is incorrect at some models
After running gltfpack to compress a model, AABB looks incorrect. I use a below command.

```
gltfpack -i model.glb -o compress.glb -cc
```

Below figure is before running gltfpack.

image

Below figure is after running gltfpack.

image

AABB should be equal to original when using gltfpack.

daemyung:
The morph target is displacement data (deltas). So the base primitive's AABB and the morph target's AABB should be added to calculate correctly.

Fix #503

### DIFF
--- a/gltf/gltfpack.cpp
+++ b/gltf/gltfpack.cpp
@@ -538,7 +538,7 @@ static void process(cgltf_data* data, const char* input_path, const char* output
 		size_t pi = i;
 		for (; pi < meshes.size(); ++pi)
 		{
-			const Mesh& prim = meshes[pi];
+			Mesh& prim = meshes[pi];
 
 			if (prim.skin != mesh.skin || prim.targets != mesh.targets)
 				break;
--- a/gltf/gltfpack.h
+++ b/gltf/gltfpack.h
@@ -30,6 +30,9 @@ struct Stream
 	int target; // 0 = base mesh, 1+ = morph target
 
 	std::vector<Attr> data;
+
+	float min[3];
+	float max[3];
 };
 
 struct Transform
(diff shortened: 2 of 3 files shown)


daemyung:
AABBs are calculated correctly with these changes.

image

zeux:
I believe this is incorrect. glTF accessor stores min/max of displacements. If you want to convert that to absolute values, you need to add that on the application side - position min/max of morph targets can't be used as AABB bounds. See https://github.com/mrdoob/three.js/blob/dev/examples/js/loaders/GLTFLoader.js lines 3890-3930 for example.

zeux:
(note also that it's incorrect to simply add min/max, as the correct AABB depends on the morph values and if the values can be negative, you need to do something like `min -= max(abs(target.max), abs(target.min)); max += max(abs(target.max), abs(target.min));` to account for worst case displacement assuming weights are in [-1, 1] range)

daemyung:
@zeux Oh I see. So adding two AABB is application's resposibility. Could you confirm it again?

daemyung:
> (note also that it's incorrect to simply add min/max, as the correct AABB depends on the morph values and if the values can be negative, you need to do something like `min -= max(abs(target.max), abs(target.min)); max += max(abs(target.max), abs(target.min));` to account for worst case displacement assuming weights are in [-1, 1] range)

Could you give me a spec link about morph target weight range? I just assume that it will be [0, 1].

zeux:
> So adding two AABB is application's resposibility. Could you confirm it again? 

Yeah, this is correct.

> Could you give me a spec link about morph target weight range? I just assume that it will be [0, 1].

Unfortunately I believe the spec doesn't restrict the weight range in any way. The reason why I usually assume [-1, 1] is that negative weights are useful for facial animations, but really any weight can work in theory. So unfortunately in principle the only way to calculate the correct AABB is to combine the target bounds with the weights at runtime (which can be expensive / prohibitively complicated etc.).

gltfpack doesn't rely on the weights being in any particular range though, so yeah it's up to the application to figure out how to best compute the AABB.

daemyung:
@zeux Thanks. I will close this PR and related issue. I try to handle it in the application side.

daemyung:
@zeux Anyway! your meshoptimizer is very helpful to me. I can reduce asset size very much. Thanks for your effort!

Issue #158, opened 2020-06-23
Attribute-aware error metrics for simplification


fstrugar:
Hi! I'm playing with https://developer.nvidia.com/orca/amazon-lumberyard-bistro dataset and meshoptimizer and I've noticed this particular failcase related to the way the corners were authored.

For example, here is the original chair mesh:
![image](https://user-images.githubusercontent.com/23105650/85418188-5fff1400-b568-11ea-800b-daa10fc3bc22.png)

Notice the rounded corners with shared vertices. They survive the first pass of meshopt_simplify to half the number of triangles fine:
![image](https://user-images.githubusercontent.com/23105650/85418376-9f2d6500-b568-11ea-9891-c565d963b2af.png)

However, once the triangles between two sides facing at 90deg get folded and the sides start sharing the vertices, the vertex normals can no longer be correct:
![image](https://user-images.githubusercontent.com/23105650/85418724-09dea080-b569-11ea-8bfe-d15dd00d6055.png)

What would be a solution to (automatically) preventing this?

I was thinking of adding additional custom skip code in the 'pickEdgeCollapses' loop if angle between vertex normals is above certain threshold but I'm sure there's a better/simpler solution, perhaps already there? :)

(instead of preventing collapse, could also allow it but duplicate verts so normals aren't shared?)

Thanks for the great library!!

zeux:
Yeah, so there's a few ways to fix this.

One is to discard and recompute normals post-simplification, possibly splitting vertices when the crease angle is too sharp. This works around the problem in a way, but of course it's not very convenient.

Another one is to factor the normal delta into the simplification as an extra error. This can be done by comparing the normals alongside the edge that's considered for collapse, or it can be done by introducing normal into the quadric weight. It's on my list to experiment more with this, there's a simplify-attr branch in this repository from my last attempt but when I worked on this at the time it became clear that this isn't very simple so I decided to take a break and think about this more.

This isn't implemented right now though, and it's definitely good to address this but I'm not sure what the best solution is, since ideally it's not just the normals that need to be taken into account, and balancing ease of use, performance and quality here is tricky...

fire:
I also encountered this problem from the 
![image](https://user-images.githubusercontent.com/32321/103106022-eadbcc80-45e6-11eb-95ea-9e6537a6bdc7.png) mesh in https://github.com/zeux/meshoptimizer/issues/206#issuecomment-748478183

fire:
[...]

PR #524, opened 2022-12-29
simplify: Initial support for attribute-aware simplification

zeux:
Note: this PR is probably going to be in draft for a while, but it's a little bit easier to track remaining work with a PR so I'll keep this open while the work is ongoing.

This currently contains a rebase of prior attribute metrics work from 2019; Godot engine currently uses a version of this change with tweaks around distance-based metric composition with attributes. There's a fair amount of outstanding work here including incorporating that and not having to hardcode the max attribute count at compile time.

Contributes to #158.


zeux:
Just a note that this has been going slow partly because I ran into some problems with quadric math as applied to weighting attribute errors. Will need a little bit of further research...

zeux:
This is getting impossibly long :sweat_smile: I'm not happy with some of the results and some of the internals, but I think I'll look into merging it in a state where it has known deficiencies and iterating on it in master instead of keeping a very long lived branch.

The issues that still remain that this might get merged without:

- The attribute metric is not perfect - it's functioning correctly and is numerically stable, but it misses certain obvious visual errors. This is the "math" issue I was referring to above, and I don't currently know how to correct this.
- The attribute quadrics are not properly aggregated across discontinuities. This is the case for Godot's fork as well. This is easy to fix, but I might fix this after merging the PR.
- The resulting error, as well as error limit, include the attribute error. Godot's fork adjusts output error to only track distance, but keeps error limit as is. I might instead add a second output error parameter. This also requires tracking both errors, which increases the collapse list structure if done naively.

There's also some nice-to-have things like incorporating attribute support into JS demo. Overall I'm tempted to solve all of these in followup changes instead of keeping this PR alive for much longer.

zeux:
After all of the work in this PR the change has no impact on behavior or performance of `meshopt_simplify`; it increases JS simplifier by ~1 KB post-gzip (5.6 => 6.8 KB) which is completely fine. So there's no downside to merging this other than the incomplete implementation of the new algorithm.

PR #538, opened 2023-02-20
meshopt_optimizeVertexFetch() typo?

garettbass:
I think the destination argument for `meshopt_optimizeVertexFetch()` should be the index buffer to rewrite.


garettbass:
My mistake, I misread the function declaration!

PR #544, opened 2023-03-21
gltfpack: Fix optimizeMaterials incorrectly optimizing on MR fallback

rafern:
If a material has both a metallic-roughness model and a specular-glossiness model, optimizeMaterials only checks the metallic-roughness model, which makes it incorrectly set the alphaMode to OPAQUE. This PR fixes the bug by checking both shading models instead of the first available model.


zeux:
Thanks!

PR #546, opened 2023-04-11
Allows both "wasm simd"  in meshtopt and "simd intrinsics" elsewhere

Kuranes:
If compiling a project with wasm simd and the auto vectorization flags of compiler it broke compilation

As emscripten also support  simd intrinsics compilation to wasm simd, the current set of flags are not compatible, ending breaking compilation.

(when compiling meshoptimizer as part of a bigger codebase, not when compiling it as a library)

### DIFF
--- a/src/vertexcodec.cpp
+++ b/src/vertexcodec.cpp
@@ -44,6 +44,10 @@
 // When targeting Wasm SIMD we can't use runtime cpuid checks so we unconditionally enable SIMD
 #if defined(__wasm_simd128__)
 #define SIMD_WASM
+// Prevent compiling other variant when wasm simd compilation is active
+#undef SIMD_NEON
+#undef SIMD_SSE
+#undef SIMD_AVX
 #endif
 
 #ifndef SIMD_TARGET
@@ -92,13 +96,7 @@
 #pragma clang diagnostic ignored "-W#pragma-messages"
 // byte, enc, encv, data_var
 #pragma clang diagnostic ignored "-Wunused-variable"
-// decodeVertexBlock
-#pragma clang diagnostic ignored "-Wunused-function"
 #include <wasm_simd128.h>
-// Prevent compiling other variant when wasm simd compilation is active
-#undef SIMD_NEON
-#undef SIMD_SSE
-#undef SIMD_AVX
 #endif
 
 #ifdef SIMD_WASM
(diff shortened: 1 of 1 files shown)


src/vertexcodec.cpp
```diff
@@ -83,9 +83,22 @@
...
 #undef __DEPRECATED
 #pragma clang diagnostic ignored "-Wdeprecated-declarations"
+#pragma clang diagnostic ignored "-W#pragma-messages"
+// byte, enc, encv, data_var
+#pragma clang diagnostic ignored "-Wunused-variable"
+// decodeVertexBlock
+#pragma clang diagnostic ignored "-Wunused-function"
 #include <wasm_simd128.h>
+// Prevent compiling other variant when wasm simd compilation is active
+#undef SIMD_NEON
```
zeux, line 99: These should go next to #define WASM_SIMD - by now it's a bit too late as we've already included headers and such

src/vertexcodec.cpp
```diff
@@ -83,9 +83,22 @@
...
 #ifdef SIMD_WASM
+// WASM simd updated some macros
+// ie: 'wasm_v32x4_shuffle' is deprecated
+// https://emscripten.org/docs/porting/simd.html
+// https://github.com/llvm/llvm-project/blob/main/clang/lib/Headers/wasm_simd128.h
 #undef __DEPRECATED
 #pragma clang diagnostic ignored "-Wdeprecated-declarations"
+#pragma clang diagnostic ignored "-W#pragma-messages"
+// byte, enc, encv, data_var
+#pragma clang diagnostic ignored "-Wunused-variable"
```
zeux, line 94: I think this and below would go away if this included WASM in the condition (I think it's an oversight that it doesn't):

```
#if defined(SIMD_FALLBACK) || (!defined(SIMD_SSE) && !defined(SIMD_NEON) && !defined(SIMD_AVX))
static const unsigned char* decodeBytesGroup(const unsigned char* data, unsigned char* buffer, int bitslog2)
```

zeux:
Could you share the command line arguments for emcc that make the build fail, so that it's easier for me to test this?

src/vertexcodec.cpp
Kuranes: using 
```cpp
#if defined(SIMD_FALLBACK) || (!defined(SIMD_SSE) && !defined(SIMD_NEON) && !defined(SIMD_AVX) && !defined(SIMD_WASM))
static const unsigned char* decodeBytesGroup(const unsigned char* data, unsigned char* buffer, int bitslog2)
```
it allows to remove that part

```cpp
// decodeVertexBlock
#pragma clang diagnostic ignored "-Wunused-function"
```

Kuranes:
`-O3 -flto -msse  -msse2  -msse3 -msse4.1 -msse4.2 -mavx -mfpu=neon -msimd128`
My understanding is that `-msse  -msse2  -msse3 -msse4.1 -msse4.2 -mavx -mfpu=neon` enable "intrinsic translation" and  `-msimd128` enable "wasm specific intrinsics" 
(emcc -v gives here version 3.1.21, and clang 16)

zeux:
Thanks!

PR #553, opened 2023-04-26
clusterizer: Implement experimental triangle mask generator

zeux:
This change computes triangle masks for each frustum slice, following "GPU-Driven Rendering Pipelines" from SIGGRAPH 2015.

The resulting masks can be used to quickly cull individual triangles, which only requires classifying the camera position in cube frustum face in the object space and selecting the appropriate set of triangle bits.

The masks can also be used to reject the entire clusters, or entire triangle groups, by comparing all of them to 0 in task shader and/or comparing each 32-triangle group to 0 in mesh shader before emitting that triangle group.

Note: this change is not fully tested, so it's not ready (may contain errors in sign/winding/etc.; also needs demo code cleanup)

### DIFF
--- a/demo/main.cpp
+++ b/demo/main.cpp
@@ -952,8 +952,10 @@ void meshlets(const Mesh& mesh, bool scan)
 	    (endc - startc) * 1000);
 
 	std::vector<unsigned int> triangle_masks(6 * ((max_triangles + 31) / 32));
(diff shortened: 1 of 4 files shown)


zeux:
Per https://zeux.io/2023/04/28/triangle-backface-culling/ the results here are not that interesting so I'll leave this up for a bit but it looks unlikely that this will be merged.

PR #559, opened 2023-05-18
Fixed vertexfilter.cpp compilation with Emscripten

LorenRoosendaal:
There was a missing undef of SIMD_SSE when we detect we're using __wasm_simd128__ in vertexfilter.cpp, funny enough vertexcodec.cpp does do this correctly.

Either way, this change fixes the WASM target compiling.


zeux:
Ah yes, vertexcodec used to have this problem but it was recently fixed. Thanks!

PR #564, opened 2023-06-05
avoid to use reserved word 'interface'

1d10t:
`Uncaught SyntaxError: Unexpected strict mode reserved word (at library.js:50:3)`

raises while loading 'library.js' as external js module.


zeux:
Thanks!

PR #567, opened 2023-06-15
add es module of gltf/library.js

1d10t:
usage: `import * as gltfpack from 'path/to/module.js'`

### DIFF
--- a/gltf/library.js
+++ b/gltf/library.js
@@ -1,0 +1,374 @@
+// This file is part of gltfpack and is distributed under the terms of MIT License.
+
+/**
+ * Initialize the library with the Wasm module (library.wasm)
+ *
+ * @param wasm Promise with contents of library.wasm
+ *
+ * Note: this is called automatically in node.js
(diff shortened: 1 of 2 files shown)


zeux:
I forgot to comment on this, but this is a little stuck due to node.js only having ES module support in recent versions. I'm not sure if there's a way to refactor this code so that we can have a module that can be imported but still have most of the code in the same file... and I don't think I want to maintain the duplicate JS files here.

Maybe an alternative is to have some simple prepublish step that builds the module.js file from library.js before creating the package, as long as this doesn't introduce new build time dependencies (aka runs in pure node).

zeux:
I'm going to close this because I don't want to maintain two versions here so I can't merge this as is. Looking at the node.js support for ESM I would be happy to accept a change that just switches library.js to a module (fine to keep the file name); my understanding is that this requires a change in package.json to make node recognize it as a module, and should be compatible with Node 12 and later versions.

PR #530, opened 2023-01-23
Preserve custom attributes

ru-imajion:
Preserves custom attributes.

Issue #165, opened 2020-07-09
Individual mesh picking on consolidated meshes - gltfpack


sweco-sekrsv:
I saw the discussion over here:
https://github.com/KhronosGroup/glTF/issues/1699

Have you already added support for this or have plans to add support for custom attributes? To be able to store the original node index integer in a custom attribute to be able to do individual mesh picking and also fetch extras data from the consolidated meshes. I think it would be very useful for many people in the AEC/CAD/BIM industry, making these models alot more accessible in more tools.

zeux:
Support for this hasn't been added. To do this we need to establish a specific naming convention, since gltfpack needs to be able to reason about this attribute in terms of expected format, and figure out if that's something that an external application should feed into gltfpack or gltfpack should synthesize by itself.

One alternate possibility is to use EXT_mesh_gpu_instancing; early experiments on CAD-like models with that were pretty encouraging. gltfpack supports it out of the box in recent builds using `-mi` command line option. This is nicer because it is a more idiomatic way to represent large collections of objects in glTF while still keeping an ability to distinguish one object from another, although it's not as performant on scenes with many different kinds of objects.

sweco-sekrsv:
Intersesting , I'll have a look at the EXT_mesh_gpu_instancing. Do you already have an example in babylon.js using this that you can share?

What defines a mesh being the same(instance) as another in the input gltf file?

**ObjectA - a windowframe**
Extras data - attribute A
Unique material: Red
**ObjectB**
Same geometry as ObjectA
Extras data - attribute B
Unique material: Red
**ObjectC**
Same geometry as ObjectA
No extras data
Unique material: Red
**ObjectD**
Same geometry as ObjectA
No extras data
Unique material: Red
**ObjectE**
Same geometry as ObjectA
No extras data
Unique material: Blue

In the above case, would ObjectC and ObjectD become one merged mesh object or will they be instances of each other or instances of all the objects? 
Would the output file contain only two materials, red and blue even if there are 5 unique materials in the original gltf file?

Do you also instansiate subelements in a mesh?
For example if we have 1000 windowframes which beforehand have been collapsed to the same meshobject, are you looking at the subelements of the mesh to identify windowframes extract them and make them into instances? Or would that be worse performant? (they would all share the same extras data) . I guess its a balance between filesize and optimization as well.

zeux:
See attached file https://github.com/BabylonJS/Babylon.js/issues/8164 (performance issues have been fixed since)

[...]

PR #568, opened 2023-06-17
gltfpack: Preserve custom ID attributes

### LINKED ISSUE #529: gltfpack: Preserve custom vertex attributes during optimization
It would be helpful to preserve custom attributes when using gltfpack. Currently, if I pass a glb in with a custom `_ID` vertex attribute, the resulting file has that attribute stripped. Instead, I'd like to see those attributes preserved, and when simplification is applied the indexes that survive simplification should be copied to the new model.

Attached are two glbs with custom `_ID` vertex attributes.

[Archive.zip](https://github.com/zeux/meshoptimizer/files/10412771/Archive.zip)

zeux:
This change keeps attributes `_ID`, `_BATCHID` and `_FEATURE_ID_*` in tact when processing the scene. We assume that these attributes are scalar and encode them as floating-point (to avoid possible range issues with uint16).

This change is similar to #530 (without the extraneous names) but it crucially limits the supported attributes by name otherwise we can not assume that the attribute is scalar or integer, whereas with this filtering the assumption is reasonably safe.

For increased compression efficiency, when `-cc` is applied we use exponential encoding unless the IDs exceed the maximum permissible range. On some test files that use feature and vertex ids, this reduces the compressed size of ID attribute streams by ~2x.

Relates to #165
Fixes #529
Closes #530

### DIFF
--- a/gltf/gltfpack.h
+++ b/gltf/gltfpack.h
@@ -29,6 +29,8 @@ struct Stream
 	int index;
 	int target; // 0 = base mesh, 1+ = morph target
 
+	const char* custom_name; // only valid for cgltf_attribute_type_custom
+
 	std::vector<Attr> data;
 };
 
--- a/gltf/parsegltf.cpp
+++ b/gltf/parsegltf.cpp
@@ -136,6 +136,13 @@ static void fixupIndices(std::vector<unsigned int>& indices, cgltf_primitive_typ
 	}
 }
 
+static bool isIdAttribute(const char* name)
+{
+	const char* uid = strstr(name, "_ID");
+
+	return uid && (uid[3] == 0 || uid[3] == '_');
+}
+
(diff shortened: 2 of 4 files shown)


petrbroz:
Btw. we've noticed that even though `_FEATURE_ID_*` attributes are preserved, the extension that uses these ([EXT_mesh_features](https://github.com/CesiumGS/glTF/tree/3d-tiles-next/extensions/2.0/Vendor/EXT_mesh_features)) is removed from the gltfpack output. Is that intentional?

zeux:
Yes, gltfpack must implement extensions for them to be preserved, as is common with other glTF processing tools. In case of EXT_mesh_features specifically, this extensions doesn't really exist yet in that it has been proposed (there's an open PR on glTF repository) but has not been merged or ratified, so I haven't looked into what it would take to support it.

XiaoChenDan:
Whether to consider the "EXT_mesh_gpu_instancing" extension to also support custom ID attributes

PR #582, opened 2023-07-14
Update README.md to remove mentions of -te option

Louis-Simon22:
The -te flag was removed a few versions back, but it was still mentionned in the readme.


zeux:
Ah, thanks! I've removed the sentence altogether as textures are only embedded into .glb, not .bin, and it seems fine to just omit it.

PR #585, opened 2023-07-24
gltfpack: Implement support for floating-point texture coordinate quantization

### LINKED ISSUE #194: gltfpack: Improve handling of meshes with very large UV bounds
gltfpack version: 0.14 and 0.15

I convert my gltf model to gltfpack format by -tc, it will generate a new model correct. And then use KTX2Loader and GltfLoader to load the target model in Threejs(v0.122.0). 

But some textures are displayed as stretch(LEFT: before convert, RIGHT: after convert),  I don't know what's the reason for this result? what should I do to resolve it?
IMG20201109_094430
IMG20201109_095614

### LINKED ISSUE #574: An option to skip or customize UV quantization in gltfpack
Hello.
We have an application where you can reassign materials on some objects from some predefined set.
The materials to choose from are stored in gltf file, but some of them are assigned to dummy meshes, so that various tools wouldn't delete them. Also materials are tiling UV range on the target primitive is outside [0..1]. But on dummies UV range IS 0..1,
So because the range is different we must scale UV's differently on different materials, therefore we can't reassign the materials anymore. (I mean we can, but UVs will scale incorrectly)

I see 3 solutions for gltfpack
- an option to skip UV quantization entirely
- an option to use boundaries from the entire file to compute scale for UVs
- an ability to provide explicit UV boundaries per material (or per primitive)

zeux:
Similarly to position quantization, we can use quantized floating point values to represent texture coordinates. This can be useful as it trades off a little bit of memory and compression efficiency for convenience and interoperability - without the use of texture transforms it becomes easy to swap textures, reuse materials, and avoid edge cases in quantization when UV bounds are very large.

While `-vtn` option is currently not very useful as it just enforces the default, in the future we could ostensibly automatically switch between normalized integers and floating point based on some heuristics around UV range.

Fixes #194
Fixes #574

### DIFF
--- a/gltf/gltfpack.cpp
+++ b/gltf/gltfpack.cpp
@@ -508,7 +508,7 @@ static void process(cgltf_data* data, const char* input_path, const char* output
 
 		comma(json_materials);
 		append(json_materials, "{");
-		writeMaterial(json_materials, data, material, settings.quantize && !settings.pos_float ? &qp : NULL, settings.quantize ? &qt_materials[i] : NULL);
+		writeMaterial(json_materials, data, material, settings.quantize && !settings.pos_float ? &qp : NULL, settings.quantize && !settings.tex_float ? &qt_materials[i] : NULL);
 		if (settings.keep_extras)
 			writeExtras(json_materials, material.extras);
 		append(json_materials, "}");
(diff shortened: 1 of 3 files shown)


zeux:
Example impact on file size on DragonAttenuation.gltf (note that texcoord data is a little under a third of the total geometry size on the model when using compression):

- default: 32 bits/texcoord
- `-c`: 18.6 bits/texcoord
- `-cc`: 18.6 bits/texcoord
- `-vtf`: 64 bits/texcoord
- `-vtf -c`: 23.1 bits/texcoord
- `-vtf -cc`: 17.2 bits/texcoord

Similarly to `-vpf`, this is mostly recommended when used together with `-cc`, or when the file size is not very important. When compression is not used (which also indicates the in-memory size), the geometry is ~3.5 MB with default options and ~4 MB with `-vtf` on this model. With `-cc` the geometry is actually a little smaller in transmission, even though it's larger in memory.

zeux:
`encodeExpParallel3`/`encodeExpParallel2` can be replaced with `meshopt_encodeFilterExp` but this can happen separately from this PR; other than that I think this is probably ready.

PR #596, opened 2023-08-21
gltfpack: Enhance .obj parsing by preserving more information

### LINKED ISSUE #594: gltfpack: obj material color
Motivation:
In obj material (.mtl) files, colors and other options could be defined, but output glb model does not contain these information.

Example:
obj and mtl file:
[input_obj.zip](https://github.com/zeux/meshoptimizer/files/12390213/input_obj.zip)

converted glb file:
[glb.zip](https://github.com/zeux/meshoptimizer/files/12390231/glb.zip)

expected: a red cube
actual: a white cube

related code:
https://github.com/zeux/meshoptimizer/blob/fd0b72336a680791a53b2a41719c738e3c376fce/gltf/parseobj.cpp#L59-L86

see also this for mapping from obj material to gltf material:
https://github.com/CesiumGS/obj2gltf/blob/a1709ac1fafa5546f592d232bde0a9f64b3bbd57/lib/loadMtl.js#L119-L159

<details><summary>quick and incomplete patch</summary>
<p>

Note this does not contains the pbr texture conversion code [here](https://github.com/CesiumGS/obj2gltf/blob/a1709ac1fafa5546f592d232bde0a9f64b3bbd57/lib/loadMtl.js#L887)

```patch
diff --git a/gltf/parseobj.cpp b/gltf/parseobj.cpp
index 14228d0..e38a885 100644
--- a/gltf/parseobj.cpp
+++ b/gltf/parseobj.cpp
@@ -62,12 +62,33 @@ static cgltf_data* parseSceneObj(fastObjMesh* obj)
                fastObjMaterial& om = obj->materials[mi];

                gm.has_pbr_metallic_roughness = true;
-               gm.pbr_metallic_roughness.base_color_factor[0] = 1.0f;
-               gm.pbr_metallic_roughness.base_color_factor[1] = 1.0f;
-               gm.pbr_metallic_roughness.base_color_factor[2] = 1.0f;
+               // baseColorFactor = material.diffuseColor;
+               gm.pbr_metallic_roughness.base_color_factor[0] = om.Kd[0];
+               gm.pbr_metallic_roughness.base_color_factor[1] = om.Kd[1];
+               gm.pbr_metallic_roughness.base_color_factor[2] = om.Kd[2];
                gm.pbr_metallic_roughness.base_color_factor[3] = 1.0f;
-               gm.pbr_metallic_roughness.metallic_factor = 0.0f;
-               gm.pbr_metallic_roughness.roughness_factor = 1.0f;
+               // emissiveFactor = material.emissiveColor
+               gm.emissive_factor[0] = om.Ke[0];
+               gm.emissive_factor[1] = om.Ke[1];
+               gm.emissive_factor[2] = om.Ke[2];
+               // Translate the blinn-phong model to the pbr metallic-roughness model
+               // Roughness factor is a combination of specular intensity and shininess
+               // Metallic factor is 0.0
+               // Textures are not converted for now
+               float specularIntensity = om.Ks[0] * 0.2125f + om.Ks[1] * 0.7154f + om.Ks[2] * 0.0721f;
+
+               // metallicFactor = material.specularColor[0]
+               gm.pbr_metallic_roughness.metallic_factor = 0.f;
+               // roughnessFactor = material.specularShininess
+               gm.pbr_metallic_roughness.roughness_factor = std::max(0.f, std::min(1.f - (om.Ns / 1000.f), 1.f));
+               // Low specular intensity values should produce a rough material even if shininess is high.
+               if (specularIntensity < 0.1) {
+                       gm.pbr_metallic_roughness.roughness_factor *= 1.0 - specularIntensity;
+               }
+               bool transparent = false;
+               if (om.d < 1.0f) {
+                       transparent = true;
+               }

                gm.alpha_cutoff = 0.5f;

@@ -77,6 +98,10 @@ static cgltf_data* parseSceneObj(fastObjMesh* obj)
                        gm.pbr_metallic_roughness.base_color_texture.scale = 1.0f;

                        gm.alpha_mode = (om.illum == 4 || om.illum == 6 || om.illum == 7 || om.illum == 9) ? cgltf_alpha_mode_mask : cgltf_alpha_mode_opaque;
+               } else {
+                       gm.pbr_metallic_roughness.base_color_factor[3] = om.d;
+                       gm.alpha_mode = transparent ? cgltf_alpha_mode_blend : cgltf_alpha_mode_opaque;
+                       gm.double_sided = transparent;
                }

                if (om.map_d.name)

```

</p>
</details>

### LINKED ISSUE #595: gltfpack: keep obj hierarchy and name
Motivation:
Currently gltfpack flattens and merges all obj meshes with the same material, so mesh hierarchy and name got lost even if `-kn` is passed.

Proposal:
With `-kn` argument, keep hierarchy and name from obj meshes, and do not merge it.
With `-km` argument, keep name from obj materials, and do not merge named obj materials.

Alternative:
[obj2gltf](https://github.com/CesiumGS/obj2gltf/) keeps hierarchy and name by default.

zeux:
When parsing .obj, gltfpack now:

- Preserves material names
- Preserves object names by parsing individual objects separately and assigning them to separate nodes
- Preserves Kd/d material parameters
- Fixes texture paths if they contain backslashes in the source .mtl
- Assigns the default scene in the output .glb file

Before merging:

- [x] Review and test more thoroughly
- [x] Measure memory/performance overhead of object split/merge on a few large scenes

Fixes #594 
Fixes #595

### DIFF
--- a/gltf/gltfpack.h
+++ b/gltf/gltfpack.h
@@ -11,6 +11,10 @@
 #define _CRT_SECURE_NO_WARNINGS
 #endif
 
+#ifndef _CRT_NONSTDC_NO_WARNINGS
+#define _CRT_NONSTDC_NO_WARNINGS
+#endif
+
 #include "../extern/cgltf.h"
 
 #include <assert.h>
--- a/gltf/parseobj.cpp
+++ b/gltf/parseobj.cpp
@@ -118,6 +118,8 @@ static void parseNodesObj(fastObjMesh* obj, cgltf_data* data)
 	data->scenes = (cgltf_scene*)calloc(1, sizeof(cgltf_scene));
 	data->scenes_count = 1;
 
+	data->scene = data->scenes;
+
 	data->scenes->nodes = (cgltf_node**)calloc(obj->object_count, sizeof(cgltf_node*));
 	data->scenes->nodes_count = obj->object_count;
 
--- a/gltf/parseobj.cpp
+++ b/gltf/parseobj.cpp
@@ -61,6 +61,9 @@ static cgltf_data* parseSceneObj(fastObjMesh* obj)
 		const fastObjMaterial& om = obj->materials[mi];
 		cgltf_material& gm = data->materials[mi];
 
+		if (om.name)
+			gm.name = strdup(om.name);
+
 		gm.has_pbr_metallic_roughness = true;
 
 		gm.pbr_metallic_roughness.base_color_factor[0] = om.Kd[0];
--- a/gltf/parseobj.cpp
+++ b/gltf/parseobj.cpp
@@ -58,14 +58,16 @@ static cgltf_data* parseSceneObj(fastObjMesh* obj)
 
 	for (unsigned int mi = 0; mi < obj->material_count; ++mi)
 	{
+		const fastObjMaterial& om = obj->materials[mi];
 		cgltf_material& gm = data->materials[mi];
-		fastObjMaterial& om = obj->materials[mi];
 
 		gm.has_pbr_metallic_roughness = true;
-		gm.pbr_metallic_roughness.base_color_factor[0] = 1.0f;
-		gm.pbr_metallic_roughness.base_color_factor[1] = 1.0f;
-		gm.pbr_metallic_roughness.base_color_factor[2] = 1.0f;
-		gm.pbr_metallic_roughness.base_color_factor[3] = 1.0f;
+
+		gm.pbr_metallic_roughness.base_color_factor[0] = om.Kd[0];
+		gm.pbr_metallic_roughness.base_color_factor[1] = om.Kd[1];
+		gm.pbr_metallic_roughness.base_color_factor[2] = om.Kd[2];
+		gm.pbr_metallic_roughness.base_color_factor[3] = om.d;
+
 		gm.pbr_metallic_roughness.metallic_factor = 0.0f;
 		gm.pbr_metallic_roughness.roughness_factor = 1.0f;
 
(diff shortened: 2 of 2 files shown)


zeux:
Looks like some .obj exporters use objects and some use groups to split the data :-/ this doesn't affect correctness of conversion but does mean this PR won't handle some input scenes depending on the exporter that was used.

kzhsw:
Also consider the roughness factor [here](https://github.com/CesiumGS/obj2gltf/blob/a1709ac1fafa5546f592d232bde0a9f64b3bbd57/lib/loadMtl.js#L962-L988)

zeux:
The textures in .obj are typically using the spec-gloss model. They could be emitted via KHR_materials_pbrSpecularGlossiness, however this extension is deprecated in favor of KHR_materials_specular I think - either way this requires a fair amount of special cased work which I don't think makes sense to do absent specific important use cases. Similarly I don't want to just tackle the roughness factor without considering the rest, so I'd rather keep the model simple (just a color + texture) for now. I would not be surprised if obj2gltf converts textures from spec+gloss to metal+roughness inferring the individual pixel values - this is very much out of scope of gltfpack.

Inferring double-sidedness based on transparency is easy but kinda specific... I'm not exactly sure what the motivation there is.

kzhsw:
The emissiveTexture and normalTexture [here](https://github.com/CesiumGS/obj2gltf/blob/a1709ac1fafa5546f592d232bde0a9f64b3bbd57/lib/loadMtl.js#L880) does not seems to be part of the spec-gloss model, could this be added?

zeux:
Yeah these should work without issues I believe as they fit as is within any model even without the use of extensions, I'll add them.

zeux:
Hmm, actually, obj2gltf seems to assume map_bump in .mtl files refers to normal maps, but Sponza example on https://casual-effects.com/data/ at least assumes map_bump refers to actual bump maps (containing height data). Unsure if obj2gltf is correct here, or what the spec says if any.

zeux:
Oh boy :-/ https://paulbourke.net/dataformats/mtl/ says:

> During rendering, the map_Kd value is multiplied by the Kd value.

... which is consistent with glTF spec. But of course, some files I am looking at have Kd at 0 and map_Kd set to a texture, so the effect of this change is that they render as black. obj2gltf ignores Kd when map_Kd is present which looks like the correct course of action in absence of an official spec.

zeux:
Given the amount of irregularities in .mtl I think this is a reasonable stopping point. Future additions will need to be evaluated on a case-by-case basis. map_bump has the aforementioned ambiguity that makes me think that absent a thorough analysis of whether one variant is common place it's not a good idea, and I wasn't able to find reasonable examples for emissive so for now both are not implemented.

PR #602, opened 2023-09-24
gltfpack: Merge identical texture objects together

zeux:
Blender glTF exporter (and possibly others) output duplicate texture objects in some exports. This blocks material merging which can result in more draw calls than necessary, and in some loaders (such as Three.JS) may result in extra texture decoding during loading and extra memory spent.

For simplicity we still assume that images and samplers in the source document are unique, but deduplicate and remove redundant texture objects now.

The process is a little different from how materials are deduplicated, as we need to maintain remap for unused textures to avoid traversing all texture views again as they are part of a complicated material structure.

### DIFF
--- a/gltf/gltfpack.cpp
+++ b/gltf/gltfpack.cpp
@@ -181,7 +181,7 @@ static void printImageStats(const std::vector<BufferView>& views, TextureKind ki
 		printf("stats: image %s: %d bytes in %d images\n", name, int(bytes), int(count));
 }
 
-static bool printReport(const char* path, cgltf_data* data, const std::vector<BufferView>& views, const std::vector<Mesh>& meshes, size_t node_count, size_t mesh_count, size_t material_count, size_t animation_count, size_t json_size, size_t bin_size)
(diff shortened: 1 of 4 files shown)


zeux:
Looks like there's no memory savings because three.js uses uri+sampler as a cache key, so this only affects material merge optimizations (or other non-three.js renderers?).

donmccurdy:
@zeux in case it's relevant — I currently merge identical texture objects in glTF Transform, but have been asked _not_ to do that in certain cases where the material uses different parts of the same texture for different purposes, e.g. color vs. normal map. In that situation three.js has difficulty sharing the texture resource while associating it with sRGB and non-sRGB data.

I'm inclined to continue merging textures, and to consider mixed colorspaces a pathological case, but I would be interested if you have any thoughts on what the glTF spec should recommend:

- https://github.com/KhronosGroup/glTF/issues/2308

donmccurdy:
That all lines up with what I've experienced, yes! Trying to make reasonable choices about a texture associated with multiple color spaces was more than I wanted to deal with though hah - hoping that logging a warning could be enough.

zeux:
Yeah I looked at the motivating discussion in glTF-Transform and that asset just feels broken so absent a clear use case with a proper asset it feels like a pathological case that doesn't need to be supported.

PR #607, opened 2023-10-01
README.md: Fix recommended meshlet size setting for 'max_triangles'

Nyran:
In their Article [Introduction to Turing Mesh Shaders](https://developer.nvidia.com/blog/introduction-turing-mesh-shaders/) Christoph Kubisch (NVIDIA) recommends a 'max_triangles' size setting for meshlets which maximizes multiples of 128 bytes. In his explanation, he points out that, since the triangle count itself (4 byte) need to be stored, number of triangles which maximize the memory are for instance 84 (sizeof(uint32_t) + 3 * 84 =256)  or 12**6** (sizeof(uint32_t) + 3 * 126 = 382 byte, which is just 2 byte less than the 3 * 128 byte which would be allocated as opposed to 8 byte using 124 as limit).

This is of course just a minor fix of the documentation which in practice will have a negligible effect in terms of rendering performance increases, however, is true to the original documentation and I think it does not hurt to change it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -303,7 +303,7 @@ To generate meshlet data, this library provides two algorithms - `meshopt_buildM
 
 ```c++
 const size_t max_vertices = 64;
-const size_t max_triangles = 124;
+const size_t max_triangles = 126;
 const float cone_weight = 0.0f;
 
 size_t max_meshlets = meshopt_buildMeshletsBound(indices.size(), max_vertices, max_triangles);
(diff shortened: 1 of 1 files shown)


zeux:
124 is not a typo. `meshopt_buildMeshlets` requires that `max_triangles` is divisible by 4; this is because each meshlet's index buffer is aligned to a 4 byte boundary, and if `max_triangles` isn't divisible by 4, the allocated storage requirements for this function become more error-prone / difficult to compute as a function of meshlet count.

This constant is taken from the demo source which has an additional comment, which might be useful to replicate.

```
	const size_t max_triangles = 124; // NVidia-recommended 126, rounded down to a multiple of 4
```

Additionally, the 64v / 126t suggested in that article is not a very strict bound in the sense that it should be extremely rare for a 64-vertex meshlet to reach 126 triangles; in absence of degenerate triangles or other topological oddiites, a regular mesh patch that you'd expect to get is around 98 triangles (7x7 quad grid). It's possible for individual meshlets to get more triangles (by taking away triangles from other meshlets, making them less optimal), and it does happen occasionally, but on meshoptimizer test suite only around 0.4% meshlets seem to max out the triangle count, so losing an extra 2 triangles should not measurably affect the resulting efficiency.

Nyran:
I see - I missed the comment in the demo source code.

Thank you for your time and the explanation!

zeux:
For posterity, average triangle/meshlet counts for a subset of my test data that has >1000 triangles to back up the analysis above (this is with v=64 t=124 from the demo)

image

Issue #605, opened 2023-09-29
gltfpack: Keeping external textures in .glb


petrbroz:
When optimizing gltfs with `gltfpack`, is it somehow possible to keep textures external, i.e., not bundle them into the output glbs? And if not, would it be possible to add a flag for this? I believe this would be really helpful when `gltfpack` is used for optimizing gltf/glb content of individual tiles in a 3D Tiles dataset where many tiles might be sharing the same textures.

zeux:
That currently happens automatically when the output is .gltf (not .glb).

petrbroz:
Thanks @zeux. Do you think it would make sense to make this kind of behavior available for glbs as well? I understand that the main purpose of glbs is to have everything bundled in one file but for the 3D Tiles use case I mentioned above, keeping the textures separate even for glbs might be useful IMHO.

zeux:
It’s certainly easy to add an option, but I’d like to make sure it’s useful. For 3D tiles, are both gltf and glb supported? Is gltf worse in some way?

Quick background: gltfpack used to default to separate textures regardless of the format and provided an option, -te, to embed textures into .bin (gltf) / .glb. This was very error prone when producing glb as separate textures there are very rare and the general expectation is that glbs are self contained. As such it was removed in favor of format dependent behavior. Rationale is more fully described here: https://github.com/zeux/meshoptimizer/issues/178

In practical terms, I haven’t measured it in a while but my expectations were that for geometry heavy data with textures stored separately, storing geometry separately is beneficial because it allows the loaders to parse glTF file and then concurrently download geometry and textures.

zeux:
By the way, in your use case, are textures being compressed or left as they are in the source file? If they were compressed, where would you want gltfpack to put them? I'm trying to come up with behavior that would simultaneously fix a few different issues so having more specifics about this use case would be great.

petrbroz:
Thanks for looking into this! Our models aren't texture heavy so we don't compress them.

wallabyway:
We take a large single GLTF, and split it up into smaller GLB files, spatially.
However, these GLB's share a common texture.  This is where it makes sense to use ext-reference texture folder.

You can see how we modified glTFPack CLI for our purpose here:
https://github.com/zeux/meshoptimizer/compare/master...petrbroz:meshoptimizer:master

It seems to work, but it would be so much better to have actual support within your official gltfpack tool.

petrbroz:
[...]

PR #608, opened 2023-10-02
gltfpack: Implement texture copying when compression is not used

### LINKED ISSUE #487: gltfpack: produce incomplete gltf asset if source contains external images
And if we do NOT use basisu(ktx2).

```
npm install -g gltfapck
mkdir output
gltfpack -i glTF-Sample-Models/2.0/Avocado/glTF/Avocado.gltf -o output/test.gltf
```

The texture images are not copied to `output`.

zeux:
When compression is not used and the output is a .gltf file, we currently simply preserve the image URIs and expect the caller to somehow copy the images. This is convenient for testing, but results in glTF files that are missing the image resources.

With this change, the newly added option -tcp uses the flow similar to the one we use for compression to write out copies of image assets when requested.

Fixes #487

### DIFF
--- a/gltf/gltfpack.cpp
+++ b/gltf/gltfpack.cpp
@@ -486,7 +486,7 @@ static void process(cgltf_data* data, const char* input_path, const char* output
 		}
 		else
 		{
-			writeImage(json_images, views, image, images[i], i, input_path, settings);
+			writeImage(json_images, views, image, images[i], i, input_path, output_path, settings);
 		}
 		append(json_images, "}");
 	}
@@ -1347,6 +1347,10 @@ int main(int argc, char** argv)
 		else if (strcmp(arg, "-tfy") == 0)
 		{
 			settings.texture_flipy = true;
+		}
+		else if (strcmp(arg, "-tcp") == 0)
+		{
+			settings.texture_copy = true;
 		}
 		else if (strcmp(arg, "-tj") == 0 && i + 1 < argc && isdigit(argv[i + 1][0]))
 		{
--- a/gltf/gltfpack.h
+++ b/gltf/gltfpack.h
@@ -141,6 +141,7 @@ struct Settings
 
 	bool texture_ktx2;
 	bool texture_embed;
+	bool texture_copy;
 
 	bool texture_pow2;
 	bool texture_flipy;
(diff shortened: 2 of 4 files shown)


zeux:
Note: this is a do-no-harm change in the sense that it's strictly opt-in, but I'm having second thoughts about the gltfpack behavior here. Other glTF processing tools don't have an option to not copy the images and just do the right thing by default; without copying the behavior when using selective texture compression (e.g. `-tc color`) is fairly convoluted and it's non-trivial to fix it up on the call side.

The original behavior comes from the time when gltfpack wasn't really aware of textures (it didn't have texture compression, and was purely a geometry processing tool). It is convenient for testing since it reduces the amount of file I/O, but I'm not actually sure that it's useful - which is to say, I'm not sure whether anyone actually *wants* the current gltfpack behavior when producing .gltf files.

I'm going to keep this PR open for now and see if there's any feedback or if I can figure out a more sensible set of rules here. 

One advantage of the status quo is that it works better when glTF assets share a texture repository. However, when texture compression is used, this use case is not supported properly as the textures always get saved to the output file path. Additionally this is sophisticated enough that it should probably be opt-in somehow, not the default, but I have little visibility into how gltfpack is used so it's unclear if this is even common.

zeux:
There's also a possible interaction with #605: if gltfpack defaulted to copying textures, and instead had an option to use the current semantics where image URIs are preserved as is, then maybe this would fix #605 as well. Of course if image compression is used, gltfpack needs to produce new files, so all bets are off again - unless in presence of this option, image compression places the compressed assets side-by-side with the source assets instead of into the open path...

So a version of this could be:

- Change default behavior to always copy image files to output folder (.gltf) or embed them (.glb), whether or not compression is used
- A separate option, name TBD, changes the behavior to keep image files in place, compress images side by side with sources, and keep URIs as is (modulo extension change when compression is used). This would be useful for sharing texture folders but would be an opt-in.

Also this change as implemented currently only keeps the texture file name; since it's common in the ecosystem to store textures in the `textures` folder, we'd probably want to try to preserve that, although this may have gotchas of its own...

zeux:
Thinking more about this, I do think the current behavior is sufficiently odd that it deserves to be changed. The trick is to try to change it with the minimum amount of extra options to cover various odd cases :) But I think just a single new option could simultaneously allow going back to the old behavior and satisfy #605, along the lines of the previous comment.

I'm going to close this PR, continue working in the associated branch and when that seems ready, will open a new PR with a more complete justification so that there's a coherent explanation to go back to for the purpose of documenting the change / release notes etc.

PR #610, opened 2023-10-04
js: Update decoder module to be more friendly to Terser

zeux:
Terser minifies and mangles JS; since we build WebWorker JS code dynamically as a string, references to identifiers from that string end up being invalid.

To reduce fragility of this code, we no longer rely on module-scoped variables instance and ready, and instead communicate the instance through self.ready promise.

See #609.

### DIFF
--- a/js/meshopt_decoder.js
+++ b/js/meshopt_decoder.js
@@ -41,7 +41,7 @@ var MeshoptDecoder = (function() {
 		return result.buffer.slice(0, write);
 	}
 
-	function decode(fun, target, count, size, source, filter) {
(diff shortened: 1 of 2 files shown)


andreasplesch:
Thanks ! I had not gotten to the point to realize that "instance" and "ready" are also used in the module scope and in the code string.

I tested the new meshopt_decoder.js from the js-terser branch and it worked fine after terser mangling for us ( https://github.com/x3dom/x3dom ) in a test branch, with the default terser options.

The use of `workerProcess.name` may look a little cryptic but github blame would point back to here to explain if comments are to be kept at a minimum.

PR #611, opened 2023-10-05
gltfpack: Improve texture file handling behavior

### LINKED ISSUE #487: gltfpack: produce incomplete gltf asset if source contains external images

### LINKED ISSUE #605: gltfpack: Keeping external textures in .glb

zeux:
Previously, when targeting .gltf files without texture compression, gltfpack left texture URIs untouched in the file. This meant that the users were required to copy the texture files by hand to the output folder so that the textures could be discovered (or somehow reference the original folder in the loader).

When using texture compression, gltfpack would instead save the files into the output folder. This meant that the output glTF file was complete, however when some textures were compressed and some weren't using selective (slot-specific) compression, this still meant that some files needed to be copied by hand.

When targeting .glb files, all textures we embedded in the output, compression or not. So the only way to guarantee complete glTF output was to use .glb destination... however, this meant that for .glb outputs textures could not be kept external, which could be a problem if the textures are shared between separate .glb files.

This change reworks the flow so that by default we guarantee that the output asset is complete, by copying files to the output folder (when target asset type is .gltf) or embedding them (when target asset type is .glb). Compression will just result in a different output file extension for textures that are compressed.

A new option, `-tr`, allows to keep the original paths without copying/embedding. When target asset type is `.gltf`, specifying `-tr` allows to revert to the old behavior. When target asset type is `.glb`, specifying `-tr` effectively keeps the textures unembedded (relying on the user to copy them). So the default behavior is now more correct, and `-tr` exists to satisfy special cases.

When texture compression is enabled, the meaning of `-tr` is unclear. While we could support it by, for example, compressing the files next to the original files, for now we explicitly forbid the combination - if this gets requested, it should be easy to add.

Notably, we currently use the same logic for determining target image paths as we do for compressed textures. This means we inherit the (potentially problematic) behavior of not preserving the relative texture paths, for example `textures/Material_1.png` ends up as `Material_1.png`. This can be improved separately, but this change will detect and flag the problem.

Fixes #487 (gltfpack now does the right thing)
Fixes #605 (using `-tr` allows to output a `.glb` asset with external textures)

### DIFF
--- a/gltf/gltfpack.cpp
+++ b/gltf/gltfpack.cpp
@@ -1348,9 +1348,9 @@ int main(int argc, char** argv)
 		{
 			settings.texture_flipy = true;
 		}
-		else if (strcmp(arg, "-tcp") == 0)
+		else if (strcmp(arg, "-tr") == 0)
 		{
-			settings.texture_copy = true;
+			settings.texture_ref = true;
 		}
 		else if (strcmp(arg, "-tj") == 0 && i + 1 < argc && isdigit(argv[i + 1][0]))
 		{
@@ -1459,6 +1459,7 @@ int main(int argc, char** argv)
 			fprintf(stderr, "\t-tp: resize textures to nearest power of 2 to conform to WebGL1 restrictions\n");
 			fprintf(stderr, "\t-tfy: flip textures along Y axis during BasisU supercompression\n");
 			fprintf(stderr, "\t-tj N: use N threads when compressing textures\n");
+			fprintf(stderr, "\t-tr: keep referring to original texture paths instead of copying/embedding images\n");
 			fprintf(stderr, "\tTexture classes:\n");
 			fprintf(stderr, "\t-tc C: use ETC1S when encoding textures of class C\n");
 			fprintf(stderr, "\t-tu C: use UASTC when encoding textures of class C\n");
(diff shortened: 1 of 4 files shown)


zeux:
(since this is a significant behavior change I'll keep it open for a week or so and do more testing and/or tweaks in the mean time)

zeux:
I've tested this more extensively over the last few days and this seems like a strict improvement per reasoning above. Because the new `-tr` option allows to revert back to previous behavior with the only case that's not 100% covered being the use of partial texture compression (which currently enforces copying even for uncompressed textures as `-tr` is unavailable), this should be safe to merge.

PR #614, opened 2023-10-07
Add coverage build for gltfpack

zeux:
For now we only run with `-cc`; this means that some paths will not be adequately covered. Additionally, some extensions may not be covered if glTF-Sample-Assets has no examples with them in use.


zeux:
This currently adds ~30 seconds to one of the longer builds (gltfpack), so might need to be split into a separate build target.

zeux:
There's also a rather large amount of code that's not covered right now... this would reduce the overall repo coverage to 80%.

zeux:
The coverage is now reasonably high (86% for gltf/ folder, 91% overall).

Missing coverage:
- Texture file writing (will get fixed automatically with #611)
- Command line option parsing/validation (not very interesting to cover)
- Point clouds (needs test assets)
- Morph target weight animations (needs test assets)
- Custom attributes (needs test assets)
- Some special cases for vertex attributes (high precision normals, uncompressed components, 16-bit joint indices)

PR #619, opened 2023-10-12
simplify: Add color support to simplifyPoints

zeux:
Following the new reservoir code, we can now easily accumulate colors into the reservoirs, and add the color error to the position error with a user-supplied weight. The color error is squared distance; as such, we scale it by the squared weight.

In addition, we now use color alpha as a point weight; the weighted sum will thus bias towards more opaque points.

Note that the alpha is not accounted for during point selection.

### DIFF
--- a/demo/main.cpp
+++ b/demo/main.cpp
@@ -524,7 +524,7 @@ void simplifyPoints(const Mesh& mesh, float threshold = 0.2f)
 	size_t target_vertex_count = size_t(mesh.vertices.size() * threshold);
 
 	std::vector<unsigned int> indices(target_vertex_count);
-	indices.resize(meshopt_simplifyPoints(&indices[0], &mesh.vertices[0].px, mesh.vertices.size(), sizeof(Vertex), target_vertex_count));
+	indices.resize(meshopt_simplifyPoints(&indices[0], &mesh.vertices[0].px, mesh.vertices.size(), sizeof(Vertex), NULL, 0, 0, target_vertex_count));
 
 	double end = timestamp();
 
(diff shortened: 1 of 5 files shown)


zeux:
A few options for handling alpha:

- Keep current code. This has a potential significant downside - we might select a point that has alpha that's far away from the average.
- Average alpha into the reservoir and incorporate it into the distance metric. A little more storage and computation.
- Keep the current RGB storage but store and compare premultiplied colors. A little more computation.

... one issue with all of these is that no point clouds in my test set have alpha, to the point where I'm not sure we even need to support alpha to begin with. Need to see if I can find better test assets here...

zeux:
I found a grand total of one Sketchfab asset with alpha :D but for that asset whether we support alpha or not seems completely immaterial as the alpha is relatively uniform with few deviations across the entire cloud, whereas the current weighting and potential improvements to candidate selection only matter when alpha is discontinuous.

There's also a downside wrt current implementation in that it requires 4-float colors; since alpha is rarely useful, this unnecessarily restricts the operation for glTF 3-float colors by requiring to expand the buffer to 4-floats. That said, without alpha the solution feels a little incomplete :-/

zeux:
Since we're lacking good test cases, and `meshopt_simplifyPoints` was apparently experimental with no color support and a significant quality bug for 4 years, it's probably fine to just add RGB support for now :) We can add alpha separately in the future, with a few different interface/implementation options to consider.

zeux:
Motivating example (scene from https://sketchfab.com/3d-models/little-cabin-3b9fbbd468e54227ad7ef7e74f8f479c, simplified with ratio 0.2):

Before:
Screenshot from 2023-10-12 18-20-26

After:
Screenshot from 2023-10-12 18-20-30

The colors on the simplified result are less noisy and it is more representative of the source scene.

PR #625, opened 2023-10-26
gltfpack: Implement end-to-end fuzzing

zeux:
While individual components of gltfpack like cgltf or mesh encoders have
their own dedicated fuzzers, gltfpack has a lot of code that can be
difficult to validate. We would expect that gltfpack is resistant to
invalid inputs (with the exception, perhaps, of .obj inputs where
fast_obj, the library we use, is currently not fuzz safe).

This change implements a in-memory GLB loading flow (parseGlb) and an
in-memory fuzzer; we disable all file I/O by using NULL paths, and
otherwise share all the same processing code with the exception of final
buffer writing.

This is work in progress as fuzzing issues are being discovered and fixed.


zeux:
This PR originally contained a few fixes to cgltf validation but I've extracted them into upstream PRs: https://github.com/jkuhlmann/cgltf/pull/233 https://github.com/jkuhlmann/cgltf/pull/234 https://github.com/jkuhlmann/cgltf/pull/235. Once they get merged this PR should result in a fairly clean fuzzing run: a couple minor problems still remain, and there's some alignment validation missing that results in UBSAN failures (which this change doesn't enable for fuzzing config) but things seem to generally be in a good shape, at least with the default setting set.

PR #629, opened 2023-10-31
Update function documentation to mention stream/attribute limits

### LINKED ISSUE #628: Why is meshopt_simplifyWithAttributes (silently) limited to 16 attributes?
The `meshopt_simplifyWithAttributes` seems to be limited to 16 attributes (i.e. 16 "floats") maximum. It will assert in the debug build, but corrupt the stack in a release build. That does not seem to be documented in the header.

But! I'm wondering why exactly 16 attributes. For a use case of say (float3 normal, float4 tangent, float2 uv, float4 blendWeights, float4 blendIndices), that is already 17 attributes. Could the limit be bumped up? There is `kMaxAttributes` which sounds like an easy change, just not sure if there are some negative effects of that.

zeux:
Also assert on these limits in JS simplifyWithAttributes; Wasm code compiles assertions out and this is an important limit right now.

Fixes #628 (for now)

### DIFF
--- a/js/meshopt_simplifier.js
+++ b/js/meshopt_simplifier.js
@@ -201,6 +201,7 @@ var MeshoptSimplifier = (function() {
 			assert(target_error >= 0 && target_error <= 1);
 			assert(Array.isArray(attribute_weights));
 			assert(vertex_attributes_stride >= attribute_weights.length);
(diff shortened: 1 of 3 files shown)


zeux:
I'm tempted to take another look at this because maybe the hard limit isn't even required.

PR #638, opened 2023-12-06
optimizeVertexCacheTable: redundant connection count buffer 

sciecode:
Auxiliary buffer `live_triangles` isn't required for `optimizeVertexCacheTable` variation.
As adjacency list `counts` is also decreased when necessary, and can be correctly used in every required instance.

### DIFF
--- a/src/vcacheoptimizer.cpp
+++ b/src/vcacheoptimizer.cpp
@@ -195,6 +195,9 @@ void meshopt_optimizeVertexCacheTable(unsigned int* destination, const unsigned 
 	TriangleAdjacency adjacency = {};
 	buildTriangleAdjacency(adjacency, indices, index_count, vertex_count, allocator);
 
+	// live triangle counts
+	unsigned int* live_triangles = adjacency.counts;
+
 	// emitted flags
 	unsigned char* emitted_flags = allocator.allocate<unsigned char>(face_count);
 	memset(emitted_flags, 0, face_count);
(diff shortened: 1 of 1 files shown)


src/vcacheoptimizer.cpp
```diff
@@ -195,10 +195,6 @@ void meshopt_optimizeVertexCacheTable(unsigned int* destination, const unsigned
 	TriangleAdjacency adjacency = {};
 	buildTriangleAdjacency(adjacency, indices, index_count, vertex_count, allocator);
 
-	// live triangle counts
```
zeux, line 198: I'd prefer to keep the variable but alias it with adjacency.counts (i.e. `unsigned int* live_triangles = adjacency.counts;`)

src/vcacheoptimizer.cpp
```diff
@@ -268,11 +264,6 @@ void meshopt_optimizeVertexCacheTable(unsigned int* destination, const unsigned
 		cache = cache_new, cache_new = cache_temp;
 		cache_count = cache_write > cache_size ? cache_size : cache_write;
 
-		// update live triangle counts
```
zeux, line 271: This should note that live triangle counts are updated as a byproduct of adjacency adjustment below

src/vcacheoptimizer.cpp
zeux: (the two uses below can keep using `live_triangles` then)

sciecode: Sure, will patch shortly.

zeux:
Thanks!

PR #651, opened 2024-01-25
Adds higher precision spatial sorting

pm4rtx:
Hi,

I've been working with some meshes where 10 bits of precision was not sufficient, so I added a version of `spatialSortRemap` with higher precision, so perhaps it could be helpful in upstream.

### DIFF
--- a/src/meshoptimizer.h
+++ b/src/meshoptimizer.h
@@ -538,11 +538,13 @@ MESHOPTIMIZER_API struct meshopt_Bounds meshopt_computeMeshletBounds(const unsig
  * Spatial sorter
  * Generates a remap table that can be used to reorder points for spatial locality.
  * Resulting remap table maps old vertices to new vertices and can be used in meshopt_remapVertexBuffer.
+ * When using spatialSortRemap64, the number of precision bits used to determine the order of points is doubled
  *
  * destination must contain enough space for the resulting remap table (vertex_count elements)
  * vertex_positions should have float3 position in the first 12 bytes of each vertex
  */
 MESHOPTIMIZER_API void meshopt_spatialSortRemap(unsigned int* destination, const float* vertex_positions, size_t vertex_count, size_t vertex_positions_stride);
+MESHOPTIMIZER_API void meshopt_spatialSortRemap64(unsigned int* destination, const float* vertex_positions, size_t vertex_count, size_t vertex_positions_stride);
 
 /**
  * Experimental: Spatial sorter
--- a/src/spatialorder.cpp
+++ b/src/spatialorder.cpp
@@ -21,7 +21,18 @@ inline unsigned int part1By2(unsigned int x)
 	return x;
 }
 
-static void computeOrder(unsigned int* result, const float* vertex_positions_data, size_t vertex_count, size_t vertex_positions_stride)
+inline unsigned long long int part1By2_64(unsigned long long int x)
+{
+	x &= 0x000fffffull;                          // x = ---- ---- ---- ---- ---- ---- ---- ---- ---- ---- ---- jihg fedc ba98 7654 3210
+	x = (x ^ (x << 32)) & 0x000f00000000ffffull; // x = ---- ---- ---- jihg ---- ---- ---- ---- ---- ---- ---- ---- fedc ba98 7654 3210
+	x = (x ^ (x << 16)) & 0x000f0000ff0000ffull; // x = ---- ---- ---- jihg ---- ---- ---- ---- fedc ba98 ---- ---- ---- ---- 7654 3210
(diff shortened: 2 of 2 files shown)


zeux:
Any chance you'd be able to share some meshes in question, publicly or privately, as well as the evaluation criteria for the sort order?

There's definitely an expectation that the quantization in spatialSortRemap produces the same values for points that are too close. But that previously has been considered fine as it didn't affect the actual order too greatly - as in, if the points are close enough then the order of points doesn't matter. That said, of course if the extents of the mesh are very large then maybe the quantization threshold becomes unacceptably small.

I don't love the idea of adding another API to solve this, however it might be good to solve the issue inside the function itself by automatically switching to a higher precision version under some condition. Additionally I'm curious if you need the full 60 bits of precision here, or if fewer bits suffice - eg if instead of 6 radix passes we could do 4 or 5, this could be good.

zeux:
I'm going to close this because I don't think adding a separate API is the right solution. I have an internal TODO item to look into this further, likely the right solution is some combination of careful increase of internal precision (with performance tests) and switching the implementation to use tiling + sorting instead of just a global sort, and all of this should actually be measured on some scene where it makes a difference.

pm4rtx:
Hi @zeux, I apologise for the prolonged radio silence with a reply; this PR went off my radar. 

`meshopt_spatialSortRemap` was used as a drop-in implementation of a triangle sorting for a BVH builder implementation, as the library was already used in the same project. One of the meshes that produced identical morton codes with 32-bit precision is ["roadBike"](https://casual-effects.com/g3d/data10/research/model/roadBike/roadBike.zip). The evaluation criterion for the sorting order wasn't exact. Still, it was desirable to have the number of triangles with an identical Morton code somewhat below 64-128. That was because the bottom-up BVH building algorithm was doing a nearest-neighbour search per triangle with a window that was expanding from the original 64-128 neighbours to the boundary where Morton codes start to differ, so having something like >200-500 triangles with identical Morton codes lead to a considerable expansion of search window which impacted the performance.

I'm OK with this PR being closed as I understand my use case requiring higher bit quantisation/sorting may be well outside the scope `meshopt_spatialSortRemap` was initially designed for.

zeux:
Thanks, this use case makes sense. Just to make sure, I assume you are not relying on Morton codes per se when you are post-processing the output of a spatial sort, but you have some threshold of closeness that is smaller than 0.1% of the extents of the model, so the 10-bit sort doesn't suffice as you get suboptimal bounds?

To be clear, I think it would be reasonable to improve the behavior of `meshopt_spatialSortRemap` for dense models, I just don't want to have two different versions of the function.

pm4rtx:
>  Just to make sure, I assume you are not relying on Morton codes per se when you are post-processing the output of a spatial sort, but you have some threshold of closeness that is smaller than 0.1% of the extents of the model, so the 10-bit sort doesn't suffice as you get suboptimal bounds?

Yes, there's no reliance on particular Morton code values; it's just reliance on how large the sub-sequence of identical Morton codes is. For example, let's imagine there's a sub-sequence of Morton codes 
M=[... 2, 2, 2, 3, 3, 4, 5, 5, 5, 5, 5, 5, 6 ...]

and for each triangle, there's a search radius R=3, so for the triangle `i` with Morton code `M[i] == 4`
the following candidate triangles are considered `[M[i-R]; M[i+R]]`:
M=[... 1, 1, ~~2~~, ~~3~~, ~~3~~, **4**, ~~5~~, ~~5~~, ~~5~~, 5, 5, 5, 6 ...]

Since `M[i+3] == 5`, and there are other other triangles with the Morton code `==5`, so the search window expands to 
M=[... 1, 1, ~~2~~, ~~3~~, ~~3~~, **4**, ~~5~~, ~~5~~, ~~5~~, ~~5~~, ~~5~~, ~~5~~, 6 ...]
because all the Morton codes are the same (which means they are part of the same spatial cluster), there's no reason to exclude some of them from the search, so the search window could expand significantly for dense meshes. This could be viewed as
> some threshold of closeness that is smaller than 0.1%

because the cluster of identical Morton codes is too large and there's a desire to make it smaller.

> To be clear, I think it would be reasonable to improve the behavior of meshopt_spatialSortRemap for dense models, I just don't want to have two different versions of the function.

Thank you! I did understand that from your previous reply. I just wanted to emphasize it's not a high-priority request.

zeux:
FYI: I've resubmitted the adapted version of this as https://github.com/zeux/meshoptimizer/pull/869. As I was working on some RT related code I needed to use a Morton order for BVH construction, and this PR came to mind so I resurrected it with a slightly lower precisio - to reduce the overhead for now it just uses 48 bit Morton codes (5 radix passes), which on the bike model attached to the other PR still eliminates the duplicates, but makes me feel better about just improving the original function instead of creating a new one.

Thanks for the contribution! I've tagged you as a co-author on the commit to preserve the contribution history.

PR #676, opened 2024-04-08
gltfpack: Replace images that can't be found or encoded with invalid URI

### LINKED ISSUE #675: gltfpack: Output GLTF is invalid if input scene has missing image files
When optimizing a GLTF file which has image entries referencing missing files, the exported image entries have neither their `uri` property nor their `bufferView` property defined.
[According to this JSON schema in the specs](https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#schema-reference-image), such entries are invalid.

When optimizing, gltfpack logs warnings about skipping these missing images entries to the console, which means it already is aware of this possibility.
As such, I would expect it to produce a valid GLTF file by simply copying these entries to the output file without any processing.

(related to KhronosGroup/glTF#2377)

zeux:
This change consolidates error handling for images that can't be read, written, or encoded, and uniformly emits an invalid embedded Base64 URI in these cases.

This keeps the file invalid as per glTF spec, but allows various loaders to load it. The source file in cases like this was semantically invalid so maybe this is fine.

Ideally we would probably want to remove references to the image. Unfortunately, to preserve glTF file validity, this requires transitive removal of texture references from all materials as well as image/texture array compaction...

Fixes #675

### DIFF
--- a/gltf/gltfpack.cpp
+++ b/gltf/gltfpack.cpp
@@ -480,9 +480,9 @@ static void process(cgltf_data* data, const char* input_path, const char* output
 		if (encoded_images.size() && !encoded_images[i].empty())
 		{
 			if (encoded_images[i].compare(0, 5, "error") == 0)
-				fprintf(stderr, "Warning: unable to encode image %d (%s), skipping (%s)\n", int(i), image.uri ? image.uri : "?", encoded_images[i].c_str());
+				writeImageError(json_images, "encode", int(i), image.uri, encoded_images[i].c_str());
 			else
-				writeEncodedImage(json_images, views, image, encoded_images[i], images[i], output_path, settings);
+				writeEncodedImage(json_images, views, image, encoded_images[i], images[i], i, output_path, settings);
 
 			encoded_images[i] = std::string(); // reclaim memory early
 		}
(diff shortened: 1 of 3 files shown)


zeux:
Removal of image/texture references *is* possible in theory. I think right now nothing prevents this from keeping the file valid. In addition to that change being much more invasive, I'm also worried about future extensions where you can't simply omit an image/texture references, eg `EXT_lights_image_based` has an array of 6 images, so what do we do if one fails to encode?

zeux:
The effect of this change is that files that had invalid or missing images previously failed to parse when using three.js GLTFLoader, and now cleanly load with the following error in the console:

Screenshot 2024-04-08 at 5 29 17 PM

gltf.report also loads the file but has some odd error handling for cases like this:

Screenshot 2024-04-08 at 5 30 52 PM

Cyphall:
[I asked for clarifications](https://github.com/KhronosGroup/glTF/issues/2377#issuecomment-2032417561) on the official GLTF repo about GLTF asset file validity in these cases, and the result is that an invalid URI indeed makes the GLTF asset file invalid **but** a valid URI pointing to a missing/corrupted file _is_ valid usage, so the produced GLTF asset file must still be valid in this case.

zeux:
Sure; I think semantically really all of these are invalid in the same sense; glTF spec may consider some subset of these invalid cases "valid", and glTF validator may or may not disagree (for example, glTF validator will parse JPEG/PNG headers and as such a file that isn't a valid PNG can be treated as invalid by the validator).

I'm going to merge this for now as it's better than status quo; it might be better to remove image references completely after all in cases like this but this can be done separately.

PR #677, opened 2024-04-10
gltfpack: Fixup image mime type between jpeg/png based on headers

zeux:
Some glTF files exported by non-compliant tool chains may use mimeType: image/png but use JPEG images. While these files are invalid, it would be helpful to make an effort to fix this as the mimeType information is fairly redundant.

This change uses image header patterns defined in glTF specification to override mime type even if it's defined based on the image contents; this is done in readImage so that all consumers (that are mime type dependent!) get the correct result.

### DIFF
--- a/gltf/image.cpp
+++ b/gltf/image.cpp
@@ -59,10 +59,22 @@ static bool parseDataUri(const char* uri, std::string& mime_type, std::string& r
 	return false;
 }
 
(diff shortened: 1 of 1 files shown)


arpu:
i wonder if a case for webp is needed https://github.com/KhronosGroup/glTF/blob/main/extensions/2.0/Vendor/EXT_texture_webp/README.md

PR #688, opened 2024-05-15
gltfpack: Use vertex attributes to guide simplification

### LINKED ISSUE #652: gltfpack: Simplification take in account the vertex color
Would be nice if the algorithm take in account the vertex color to preserve the shape when of differents vertex color zone, here an example : 
image
image

The mouth area is particulary destroy in this example while there is so many other triangle to collapse,

The model which have no texture is good to test, all copyright to his author from sketchfab - free download : 
https://sketchfab.com/3d-models/bananya-birbo-900ce1ec328e486296cfee70b9b13a45

zeux:
When -sv command line argument is specified, we will use vertex colors
or vertex normals as part of attribute aware simplification to produce better
results. For now the intention is to use a single argument to
incorporate multiple different types of attributes with fixed weights,
in the future we might expose more detailed control via '-sv?' prefix.

Fixes #652

### DIFF
--- a/gltf/gltfpack.cpp
+++ b/gltf/gltfpack.cpp
@@ -387,7 +387,7 @@ static void process(cgltf_data* data, const char* input_path, const char* output
 		{
 			Mesh kinds = {};
 			Mesh loops = {};
-			debugSimplify(mesh, kinds, loops, settings.simplify_debug);
+			debugSimplify(mesh, kinds, loops, settings.simplify_debug, settings.simplify_attributes);
 			debug_meshes.push_back(kinds);
 			debug_meshes.push_back(loops);
 		}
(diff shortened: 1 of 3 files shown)


zeux:
This PR currently uses *either* color *or* normal; implementing support for both will require copying attribute data out of source buffers, but is likely valuable - need to do more tests.

Normal effect, before / after:

image
image

Color effect, before / after (this model doesn't use textures):

image
image

zeux:
Left to right: only using normals, only using colors, using normals+colors (equal weight)

   

The last one is still not ideal but that's probably because the underlying attribute math needs improvements. But incorporating normals is a clear win over just colors on this model.

zeux:
I think I'm going to merge this as is for now and explore multi stream version separately, as it might also be valuable to do some preprocessing on data (eg linear->sRGB transform for colors for perceptual improvements or some such).

PR #692, opened 2024-05-22
gltfpack: Implement support for KHR_texture_basisu inputs 

### LINKED ISSUE #537: gltfpack removes existing KHR_textures_basisu extension
Running `gltfpack` on a file already using the `KHR_textures_basisu` extension removes all traces of the extension in the output file

zeux:
When the input file has KHR_texture_basisu extension, we now will
preserve the images along with the extension in the output file instead
of discarding it.

When `-tc` is specified, images like this will not be recompressed; accordingly,
some flags like `-tl` and `-tp` will not work with pre-compressed inputs.

When the input textures specify both uncompressed and compressed image,
the behavior is likely to be suboptimal...

Fixes #537.

### DIFF
--- a/gltf/README.md
+++ b/gltf/README.md
@@ -81,6 +81,7 @@ gltfpack supports most Khronos extensions and some multi-vendor extensions in th
 - KHR_materials_variants
 - KHR_materials_volume
 - KHR_mesh_quantization
+- KHR_texture_basisu
 - KHR_texture_transform
 - EXT_meshopt_compression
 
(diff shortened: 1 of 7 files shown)


zeux:
TBD:

- [ ] Automatically fix assets like AnisotropyBarnLamp: https://github.com/KhronosGroup/glTF-Sample-Assets/issues/117
- [ ] Better behavior when `-tc` isn't used and the input file has both uncompressed and compressed textures (when `-tc` is used we can't really do anything meaningful...)

PR #710, opened 2024-06-24
simplify: Make hasTriangleFlip predicate use an angle cutoff

### LINKED ISSUE #346: Flipped triangles when simplifying mesh
Hi!
First of all, thank you for working on this great tool, much appreciated! 👍 

I'm trying to simplify some terrain tiles which are basically just grids with a bit of elevation. The problem is that sometimes, the simplification process generates weird meshes, with some triangles flipped, others overlapping and some very thin. You can see all three of them in the "after.obj" attached.
Here's the code I use for simplifying the mesh, although the same happens without the index and vertex remapping, just that the buffers will be bigger. Also, I tried tweaking the threshold and targetError, with similar results.
``` c++
    // Simplify mesh
    float threshold = 0.2f;
    size_t targetIndexCount = size_t(indexes.size() * threshold);
    float targetError = 1e-2f;

    indexes.resize(meshopt_simplify(indexes.data(), indexes.data(), indexes.size(), &(positions[0].x), positions.size(), sizeof(glm::vec3), targetIndexCount, targetError, nullptr));

    std::vector<unsigned int> vertexRemap(positions.size());
    size_t newVertSize = meshopt_optimizeVertexFetchRemap(vertexRemap.data(), indexes.data(), indexes.size(), vertexRemap.size());
    vertexRemap.resize(newVertSize);

    meshopt_remapIndexBuffer(indexes.data(), indexes.data(), indexes.size(), vertexRemap.data());
    meshopt_remapVertexBuffer(positions.data(), positions.data(), positions.size(), sizeof(glm::vec3), vertexRemap.data());
    meshopt_remapVertexBuffer(texCoords.data(), texCoords.data(), texCoords.size(), sizeof(glm::vec2), vertexRemap.data());
    positions.resize(newVertSize);
    texCoords.resize(newVertSize);
```

Here you can find the `before` and `after` meshes with the fewest grid cells I could reproduce the problem with and without normals, uvs etc: [Desktop.zip](https://github.com/zeux/meshoptimizer/files/7351851/Desktop.zip)

I briefly tried to debug the problem myself, but it wasn't obvious to me what happens there. Please let me know if you need more details.

zeux:
Instead of checking if the dot product is 0 or negative (90 degree limit on triangle rotations), use a cutoff of ~75 degrees. The motivation here is that meshes with complex curvature can end up flipping triangles through a series of collapses, and since we always evaluate flips compared to the current mesh we miss that. Technically this can happen even with a 75 degree limit, but in practice this solves cases of flips that have been reported before, and doesn't noticeably constrain simplification or slow it down.

For numerical stability, we avoid dividing by the triangle areas (as they can be close to zero) and instead use the areas to compute a bound for dot product. Even if the area product underflows, we just get the same check that we used to have -- areas should be in [0..1] so there should be no risk of overflow.

Fixes #346.

### DIFF
--- a/src/simplifier.cpp
+++ b/src/simplifier.cpp
@@ -892,7 +892,13 @@ static bool hasTriangleFlip(const Vector3& a, const Vector3& b, const Vector3& c
 	Vector3 nbc = {eb.y * ec.z - eb.z * ec.y, eb.z * ec.x - eb.x * ec.z, eb.x * ec.y - eb.y * ec.x};
 	Vector3 nbd = {eb.y * ed.z - eb.z * ed.y, eb.z * ed.x - eb.x * ed.z, eb.x * ed.y - eb.y * ed.x};
 
-	return nbc.x * nbd.x + nbc.y * nbd.y + nbc.z * nbd.z <= 0;
+	float ndp = nbc.x * nbd.x + nbc.y * nbd.y + nbc.z * nbd.z;
+	float abc = nbc.x * nbc.x + nbc.y * nbc.y + nbc.z * nbc.z;
+	float abd = nbd.x * nbd.x + nbd.y * nbd.y + nbd.z * nbd.z;
+
+	// scale is cos(angle); somewhat arbitrarily set to ~75 degrees
+	// note that the "pure" check is ndp <= 0 (90 degree cutoff) but that allows flipping through a series of close-to-90 collapses
+	return ndp <= 0.25f * sqrtf(abc * abd);
 }
 
 static bool hasTriangleFlips(const EdgeAdjacency& adjacency, const Vector3* vertex_positions, const unsigned int* collapse_remap, unsigned int i0, unsigned int i1)


zeux:
Here's a comparison in terms of the evaluated collapses: for each pass, this plots the log10 of collapses we've evaluated (X) vs what percentage of the evaluations were "wasteful", as in resulted in an invalid collapse (Y). Blue points are before the change, red are after, the data set is a wide set of different meshes and scenes of varying complexity.

There is some impact here, as in we do evaluate more collapses which does result in a higher waste % sometimes, but overall the impact of this seems fairly moderate. There are improvements to the pass scheduling I am planning to investigate in the future which would probably offset some of this loss as well.

image

zeux:
P.S. I should note that while this workaround is pragmatic, I don't love it :sweat: Hopefully it will be replaced by something else in the future. But with a steady volume of reports about edge case flips and this solving them, it's a reasonable stop-gap.

RikoOphorst:
@zeux Only just seen this MR, so a bit late on the reply here - but wanted to express that your work on this is appreciated. It's great to see your attention to detail and your impact analysis on this. Looking forward to trying this out!

PR #714, opened 2024-06-28
A potential performance improvement to rankEdgeCollapses

Sluggernot:
This change passes the tests, locally. I have observed that making these changes reduces the load time of meshes in Godot by ~60 to 75%
Tests in Godot were done with several .glb files from official Godot test projects as well as randomly assorted files I had available.
I freely admit I don't understand why this would make such a large difference. I originally had a huge set of changes put together in my copy of the Godot source, in this file, attempting to inline large amounts of calculations in the quadricError functions.
I found this change entirely by accident. (And originally had it completely wrong, rewriting over the same reference for each iteration of this loop)

### DIFF
--- a/src/simplifier.cpp
+++ b/src/simplifier.cpp
@@ -1011,17 +1011,18 @@ static size_t pickEdgeCollapses(Collapse* collapses, size_t collapse_capacity, c
 
 static void rankEdgeCollapses(Collapse* collapses, size_t collapse_count, const Vector3* vertex_positions, const float* vertex_attributes, const Quadric* vertex_quadrics, const Quadric* attribute_quadrics, const QuadricGrad* attribute_gradients, size_t attribute_count, const unsigned int* remap)
 {
+	Collapse* c = &collapses[0];
 	for (size_t i = 0; i < collapse_count; ++i)
 	{
-		Collapse& c = collapses[i];
+		c = &collapses[i];
 
-		unsigned int i0 = c.v0;
-		unsigned int i1 = c.v1;
+		unsigned int i0 = c->v0;
+		unsigned int i1 = c->v1;
 
 		// most edges are bidirectional which means we need to evaluate errors for two collapses
 		// to keep this code branchless we just use the same edge for unidirectional edges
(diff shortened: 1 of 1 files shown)


Sluggernot:
Let me know what sort of proof you would like for the performance changes.

zeux:
> I have observed that making these changes reduces the load time of meshes in Godot by ~60 to 75%

Please include detailed profiling methodology - what platform have you built Godot on, what compiler are you using, what compilation options are you using, and how exactly you are measuring the time. I ran this change on gcc/linux using meshoptimizer's demo program; you can do the same by building it via make or CMake and running demo/main with flags `-d path-to-a-large-file.obj`. For example:

```
~/meshoptimizer (perf) $ make config=release dev files=data/buddha.obj
build/release/meshoptimizer -d data/buddha.obj
# data/buddha.obj: 549409 vertices, 1087474 triangles; read in 89.62 msec; indexed in 104.78 msec
Simplify : 1087474 triangles => 217494 triangles (0.01% deviation) in 463.95 msec
SimplifyN: 1087474 triangles => 224268 triangles (0.30% deviation) in 216.43 msec, clusterized in 654.77 msec
```

Cleroth:
Op removed a line in his testing. https://i.imgur.com/Rgi475N.png

Sluggernot:
WOAH! Thanks

Sluggernot:
I must need some sleep. Collapse doesnt have a distance_error data member...

zeux:
Ah I see. Godot adds `distance_error` in a patch (it's important due to an abnormally high normal weighting it is using atm); it makes sense that if that line is removed, the overall performance of the process is different - for example I think that is used to determine ray cast thresholds, and setting this to zero may result in null rays which probably makes ray casting way faster. There's some other impact on the LOD generation flow from this in Godot.

I would not recommend trying to optimize `meshopt_simplify` without a lot of care. I spent some time analyzing this and wrote a larger comment here https://github.com/godotengine/godot/issues/93587#issuecomment-2196030252.

Sluggernot:
Once again, I really appreciate your time on this. Thank you very much.

PR #717, opened 2024-07-01
Integrate Prettier for all JS source

zeux:
This makes the code style consistent across various uses of JS and also
improves integration with editors that use LSP format-on-save like Zed.

As part of this, Wasm patching has been changed to recognize trailing embed
comments as that is friendly to Prettier's line splitting.


zeux:
The test data is decidedly less readable now but it's probably still worthwhile.

zeux:
Not sure I like some aspects here; my biggest issue is printWidth behavior. It's similar to clang-format for fixed-width column limit except that when a line exceeds that, prettier breaks out every expression on a single line. This PR currently works around that by using an excessively large printWidth but that also results in new long lines being created in opposition to existing line breaks...

zeux:
Going to cherry-pick some consistency changes like tabs and builds separately but for now we probably don't really need this.

PR #720, opened 2024-07-11
simplify: Rework internal debug visualization code

zeux:
We have been relying on smuggling the kind/loop info computed during
classification via meshopt_simplifyDebug global pointers. This
information is valuable to visualize, and there's future improvements to
classification that would need to be debugged, so it's not ready to
remove but exposing this as globals results in issues as it expands the
exported API surface, making it different between release & debug builds.

Instead, this change adds a way to request this via an option bit, and
return it via indices[] - instead of per-vertex information, we return
this per-triangle-corner. This fixes all the issues with exported symbols,
and even makes the information available to JS (the size overhead is
insignificant) if we decide to use simplify.html for visualization.

The newly added option bit is *not* part of the official API surface -
the format of the data may change at any point, and the entire feature
may get dropped at any point once it stops being useful.

### DIFF
--- a/gltf/mesh.cpp
+++ b/gltf/mesh.cpp
@@ -527,8 +527,13 @@ static Stream* getStream(Mesh& mesh, cgltf_attribute_type type, int index = 0)
 	return NULL;
 }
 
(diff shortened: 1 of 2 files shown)


RikoOphorst:
@zeux Fantastic, this is very useful.

PR #742, opened 2024-08-20
simplify: Separate attribute quadrics on attribute discontinuities

zeux:
Initial implementation of attribute metric used the same indexing that
we use for positional quadrics, where multiple vertices with the same
position are collapsed into a single element. This is correct and
optimal for positions, but results in problems for attributes when the
vertex is on a seam: two associated vertices carry very different
attributes, and evaluating any attribute against a vertex from an
opposite side of the seam results in a large error. This results in a
suboptimal collapse sequence and an artificially inflated resulting
error.

This change fixes this by preserving the original indexing for attribute
quadrics and carefully aggregating these during collapses. For simplicity,
and because it does not seem crucial for quality, we still approximate
an error for one edge collapse using only one side of the seam (this
distinction only matters for seam-seam edges).

Also includes improved complex vertex handling and much more detailed
trace logging which is helpful for debugging unit tests or very small
reproducers.

Contributes to #158.

### DIFF
--- a/demo/tests.cpp
+++ b/demo/tests.cpp
@@ -1347,12 +1347,12 @@ static void simplifySeam()
 	    0, 1, 0, 1,
 	    0, 2, 0, 1,
 	    1, 0, 0, 0,
-	    1, 1, 1, 0,
-	    1, 1, 1, 1,
+	    1, 1, 0.3f, 0,
+	    1, 1, 0.3f, 1,
 	    1, 2, 0, 1,
 	    2, 0, 0, 0,
-	    2, 1, 0, 0,
-	    2, 1, 0, 1,
+	    2, 1, 0.1f, 0,
+	    2, 1, 0.1f, 1,
 	    2, 2, 0, 1,
 	    3, 0, 0, 0,
 	    3, 1, 0, 0,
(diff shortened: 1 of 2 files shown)


zeux:
As an example for why this can be valuable, with a sufficiently high normal weight on seams along the cylinder caps previously we would commonly constrain the movement of the vertices both along the seam and towards the seam because of the normal quadrics from different faces (pointing up and pointing sideways) blending together. Visible on this model: https://sketchfab.com/3d-models/ruby-rose-2270ee59d38e409491a76451f6c6ef80

where before this change the simplifier would try hard to keep these at their original tessellation, which would penalize the rest of the mesh:
image

and with this change these (small) parts of the mesh are more aggressively simplified, which improves quality in other mesh areas that are more prominent
image

zeux:
Additionally, I've done very limited testing on this but before this change, feeding texture coordinates as attributes to the simplifier would be fairly counter-productive because it would penalize the UV seams way too much when UV shells are far away from each other, making the weights effectively impossible to tune well; with this it is probably more reasonable to give simplifier access to UVs with a small weight in hopes that it will somewhat improve the texture distortion.

PR #745, opened 2024-08-22
js: Adjust code style and minor internal details in MeshoptClusterizer

zeux:
This is a followup to #738; it mostly just changes internal code style to match other JS modules, simplifies various expressions and adds assertions. Some changes (especially tests) are due to Prettier's formatting.

Notably, there is an interface change to computeMeshletBounds: it previously returned the first element of the array if there was just one meshlet. This is problematic because the caller needs to be aware of this detail to correctly handle small meshes, so this change makes it always return an array.

fyi @JolifantoBambla


zeux:
Also confirmed this is working end to end on a very quickly thrown together half-AI-generated code:

```js
	function generateMeshlets(mesh) {
		const geometry = mesh.geometry;
		const positions = geometry.attributes.position.array;
		const indices = geometry.index.array;
		const vertexCount = geometry.attributes.position.count;
		const vertexStride = 3;
		const max_vertices = 64;
		const max_triangles = 124;

		MeshoptClusterizer.ready.then(async function () {
			const buffers = MeshoptClusterizer.buildMeshlets(indices, new Float32Array(positions), vertexStride, max_vertices, max_triangles);

			for (let i = 0; i < buffers.meshletCount; i++) {
				const meshlet = MeshoptClusterizer.extractMeshlet(buffers, i);
				createMeshletMesh(meshlet, positions);
			}
		});
	}

	function createMeshletMesh(meshlet, positions) {
		const geometry = new THREE.BufferGeometry();
		const vertices = [];

		// Add positions
		for (let i = 0; i < meshlet.vertices.length; i++) {
			const vertexIndex = meshlet.vertices[i];
			vertices.push(positions[vertexIndex * 3], positions[vertexIndex * 3 + 1], positions[vertexIndex * 3 + 2]);
		}

		const indices = Array.from(meshlet.triangles);

		geometry.setAttribute('position', new THREE.Float32BufferAttribute(vertices, 3));
		geometry.setIndex(indices);

		// Random color
		const color = new THREE.Color(Math.random(), Math.random(), Math.random());

		const material = new THREE.MeshBasicMaterial({ color: color });
		const mesh = new THREE.Mesh(geometry, material);

		scene.add(mesh);
	}
```

image

PR #747, opened 2024-08-23
simplify: Minor improvements in quadricFromTriangleEdge

zeux:
Instead of normalizing p10 we can correct the projection by the edge
length; since we normalize the vector anyway we instead scale p20 as
this avoids the need to handle singularity.

This makes the function a little faster; more importantly it reduces the
average error by ~1.5x (computed by comparing sqrt(quadric error) for p2
with the double precision distance-to-line computation)


zeux:
The macOS failure is because clang is using fnmsub/fmadd for the new code which is slightly changing the tessellation of the result (both platforms return the correct result, it's just different)...

jams3223:
Wow! Valve?

PR #752, opened 2024-08-27
Increase parallelism on GitHub Actions

zeux:
This should better take advantage of the 4-core setup that GHA uses now.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -90,13 +90,13 @@ jobs:
     - name: make
       run: make -j2 config=sanitize gltfpack
     - name: test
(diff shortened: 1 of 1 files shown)


zeux:
... should, but doesn't :-/

zeux:
For posterity, the only conclusively beneficial change I ended up with is using `-P` to run gltfpack in parallel. Using `-P 4` and `-P 2` provides similar timings despite GitHub claiming that the agents have 4 cores, probably the systems are overprovisioned.

Things that didn't help: using -j 4 for make builds, using CL_MPCount=4 for MSBuild builds, building debug and release builds for Windows in parallel.

PR #758, opened 2024-09-02
gltfpack: Use quantized normal/tangents for reindexing by default

zeux:
While reindexing full precision streams is usually sufficient to get a good non-redundant index buffer, some meshes go through export or processing pipelines that introduce imprecision in individual normals or tangents, creating artificial seams. This increases the storage size and decreases rendering efficiency, in addition to creating issues for simplification.

This can be solved with a tolerance-based weld pass, but for now we can simply use quantized normals/tangents during reindexing. Note that this by itself does not reduce precision of individual vertices, so in theory we can be *more* aggressive here with quantization bits - for now we always use 8 bits which should not cause issues even if the requested normal bits is 16.

### DIFF
--- a/gltf/gltfpack.cpp
+++ b/gltf/gltfpack.cpp
@@ -393,7 +393,7 @@ static void process(cgltf_data* data, const char* input_path, const char* output
 		{
 			Mesh kinds = {};
 			Mesh loops = {};
-			debugSimplify(mesh, kinds, loops, settings.simplify_debug, settings.simplify_error, settings.simplify_attributes);
+			debugSimplify(mesh, kinds, loops, settings.simplify_debug, settings.simplify_error, settings.simplify_attributes, settings.quantize && !settings.nrm_float);
 			debug_meshes.push_back(kinds);
 			debug_meshes.push_back(loops);
 		}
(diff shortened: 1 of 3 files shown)


zeux:
This usually doesn't affect the reindexing results, but on some assets that have this issue the results can be significant:

```
input: 21 mesh primitives (1174473 triangles, 1317060 vertices); 21 draw calls (21 instances, 1174473 triangles)

before:
output: 2 mesh primitives (1174473 triangles, 1251636 vertices); 2 draw calls (2 instances, 1174473 triangles)
output: buffers: vertex 7553417 bytes, index 1939480 bytes

after:
output: 2 mesh primitives (1174473 triangles, 792117 vertices); 2 draw calls (2 instances, 1174473 triangles)
output: buffers: vertex 4802215 bytes, index 1353812 bytes
```

zeux:
Actually, testing this on glTF-Sample-Models, the issue is more prevalent than I expected. E.g. DragonAttenuation goes from 137k vertices to 98k vertices. A handful of models I've seen that have issues with this are a result of Blender export following an FBX reimport, so maybe there's some specific toolchains that introduce this problem more often than others.

In the future we might replace this with a tolerance based weld (which meshoptimizer currently doesn't expose), but quantization based welding is good enough for now (although there are cases where normals are very close but fall into different quantization buckets, these cases are comparatively rare).

zeux:
Also the effect on performance is neutral; we spend a little bit of extra time reprocessing normals/tangents, but the remap is a little faster, and ultimately other steps like vertex cache optimization dominate anyhow.

PR #760, opened 2024-09-05
demo: Initial hierarchical clustered simplification (Nanite) demo

zeux:
This PR adds a new demo file, nanite.cpp, that implements Nanite-style hierarchical simplification.
Working off the SIGGRAPH presentation by Brian Karis, it builds a DAG of clusters; for each cluster,
it tracks the bounds and error that can be used to do LOD selection in parallel / independently at
runtime.

The goal of this code is *not* to be a production quality pipeline (yet!); it is to serve as a testing
ground for future improvements in meshoptimizer's algorithms as well as development of new
algorithms. In the future this may become a good starting point, but for now the main purpose
is to be a test harness, as such there's a bunch of statistics / visualization support code here (and
more to come in the future at some point).

Also, to have a reasonable baseline, this code integrates with METIS to optionally use it to build
clusters, as well as to partition clusters. Without METIS we can still build clusters using meshopt,
but for partitioning we resolve to just sequentially merge clusters (which is obviously suboptimal).

METIS triangle partitioning is generally better in terms of maintaining contiguous clusters, but
worse in terms of generating a good cluster fill. There might be a way to coerce METIS to generate
better clusters with more tweaks; currently what often happens is that it decides to split a 256-size
partition into 127 and 129 triangles, and then has no choice but to split the 129-triangle partition
into two smaller ones. This may be fixable by repeatedly asking METIS to generate partitions with
different triangle counts but this becomes even more expensive... since METIS is only used here to
set a baseline, this is probably fine.

METIS cluster partitioning has surprising failure cases where it decides that a good partition should
consist of 4 clusters that barely connect; this was unexpected because it doesn't seem to happen during
triangle builds. Maybe with this in mind, a reasonable (unordered) greedy cluster merge could work
better, but this will need to be investigated separately in the future anyway.

The actual simplification currently uses automatic border detection; it would be better to switch to
using external vertex tagging via `vertex_lock`, but for now models I use for testing that have a lot of
room to improvement are not bottlenecked by this.

The actual DAG build is also very sensitive to all of the details above in ways that seem somewhat
difficult to control. If a group at any point earlier in the pipeline is stuck (eg can't be simplified too
much), it could be because the group has internally bad topology (faceting, UV seams), but could
also be because the group is formed of disjointed clusters, or the clusters themselves are formed
of disjointed triangles. This is bad because a) the group itself is kept at a high resolution, b) removing
this group from future simplification permanently locks the boundary edges for other clusters,
which exacerbates the problem.

To try to get around this, the DAG flow has very lax simplification limits (we keep anything where we
can remove 15% of triangles without breaking borders), and also keeps anything that gets stuck in
the queue and retries later - this in some cases makes things worse but often helps.

The error propagation is using a simpler variant of the full error described in Nanite paper; their
error looks at projected bounds whereas we just approximate it, but that doesn't matter as much.
What does matter for correctness is enforcing that the error is monotonic, which we do by ensuring
spherical group bounds are merged conservatively.

Finally, the METIS support code is written with zero regards towards efficiency; it uses bad containers
and inefficient algorithms in some places for expediency. This is because hopefully in a year or so
all of the METIS code can be removed because meshopt replacements will be superior. :crossed_fingers: 

The demo can be coerced to dump parts of the mesh as .obj via DUMP env var, which can be pretty
conveniently visualized in Blender:

image

Future work here includes more statistics around meshlet quality, maybe fixes around METIS usage
to improve the baseline a bit, and more importantly this will be used to improve the library in the future :)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -100,7 +100,7 @@ endif()
 set(TARGETS meshoptimizer)
 
 if(MESHOPT_BUILD_DEMO)
-    add_executable(demo demo/main.cpp demo/tests.cpp tools/objloader.cpp)
+    add_executable(demo demo/main.cpp demo/nanite.cpp demo/tests.cpp tools/objloader.cpp)
     target_link_libraries(demo meshoptimizer)
 endif()
 
--- a/Makefile
+++ b/Makefile
@@ -108,6 +108,9 @@ check: $(DEMO)
 dev: $(DEMO)
 	$(DEMO) -d $(files)
 
+nanite: $(DEMO)
+	$(DEMO) -n $(files)
+
 format:
 	clang-format -i $(LIBRARY_SOURCES) $(DEMO_SOURCES) $(GLTFPACK_SOURCES)
 
--- a/Makefile
+++ b/Makefile
@@ -35,6 +35,11 @@ ifdef BASISU
     endif
 endif
 
+ifdef METIS
+    $(DEMO_OBJECTS): CXXFLAGS+=-DMETIS
+    $(DEMO): LDFLAGS+=-lmetis
+endif
+
 WASMCC?=$(WASI_SDK)/bin/clang++
 WASIROOT?=$(WASI_SDK)/share/wasi-sysroot
 
--- a/demo/main.cpp
+++ b/demo/main.cpp
@@ -138,21 +138,21 @@ Mesh parseObj(const char* path, double& reindex)
 	return result;
 }
 
-void dumpObj(const Mesh& mesh, bool recomputeNormals = false)
+void dumpObj(const std::vector<Vertex>& vertices, const std::vector<unsigned int>& indices, bool recomputeNormals = false)
 {
 	std::vector<float> normals;
 
 	if (recomputeNormals)
 	{
-		normals.resize(mesh.vertices.size() * 3);
+		normals.resize(vertices.size() * 3);
 
-		for (size_t i = 0; i < mesh.indices.size(); i += 3)
+		for (size_t i = 0; i < indices.size(); i += 3)
 		{
-			unsigned int a = mesh.indices[i], b = mesh.indices[i + 1], c = mesh.indices[i + 2];
+			unsigned int a = indices[i], b = indices[i + 1], c = indices[i + 2];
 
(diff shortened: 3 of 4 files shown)


JMS55:
Btw, I highly recommend downloading a really large asset like https://quixel.com/megascans/free?category=3D%20asset&assetId=siEoZ (make sure to check the download source quality asset option) and testing on that.

JMS55:
> METIS cluster partitioning has surprising failure cases where it decides that a good partition should
consist of 4 clusters that barely connect; this was unexpected because it doesn't seem to happen during
triangle builds. Maybe with this in mind, a reasonable (unordered) greedy cluster merge could work
better, but this will need to be investigated separately in the future anyway.

I've heard that adding weights based on cluster center distances can help.

> The actual simplification currently uses automatic border detection; it would be better to switch to
using external vertex tagging via vertex_lock, but for now models I use for testing that have a lot of
room to improvement are not bottlenecked by this.

Why so? Don't you want to lock the vertices of the meshlet group border? Which is what SimplifyOptions::LockBorder should do afaik?

zeux:
> Don't you want to lock the vertices of the meshlet group border? Which is what SimplifyOptions::LockBorder should do afaik?

This *only* matters when the source asset has geometric borders. If it does, then they will be locked just as meshlet borders are, when they could in theory be simplified instead. 

For example, the asset you linked has geometric borders :sweat_smile: highlighted here:

image

As a result, using LockBorder in this case will lock that border in all clusters, and you will never get below a fairly large triangle count as a result.

This is easy to change; similarly to cluster boundary edge detection, every vertex that is shared between any two clusters can be tagged as locked before processing all clusters in the queue at a given level or a group. But since I'm testing on assets that don't have this, I'm cutting this particular corner for now.

JMS55:
Are you saying that meshopt's locked vertices are based on the vertex data, and not the index data I provide? E.g. using `SimplifyOptions::Sparse | SimplifyOptions::LockBorder` is locking the edges of the entire mesh whenever simplifying? If so, I should definitely switch to manual vertex locks... 😅

zeux:
It is based on index data. But the edge that is a border edge in the source geometry will always remain a border edge for any subset of the mesh.

PR #769, opened 2024-09-14
gltfpack: Use UTF-8 encoding during gltfpack processing

### LINKED ISSUE #767: gltfpack: UTF-8 encoding of glTF JSON on windows
Env: Windows 10 with the [UTF-8 region option](https://stackoverflow.com/questions/56419639/what-does-beta-use-unicode-utf-8-for-worldwide-language-support-actually-do) disabled by default.

Steps to reproduce:
1. get any model with non-ASCII name, e.g. `é.glb`
2. run `gltfpack -i é.glb -o é.gltf`
3. open `é.gltf` in text editor and check its encoding

Actual:
`é.gltf` is not in UTF-8 encoding

Expected:
glTF JSON is encoded in UTF-8 for whatever system encoding

Spec here: <https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#json-encoding>

> glTF JSON data SHOULD be written with UTF-8 encoding without BOM. This requirement is not applied when a glTF implementation does not control string encoding. glTF implementations SHOULD adhere to [RFC 8259](https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#json), Section 8.1. with regards to treating BOM presence.

zeux:
When main() is used, the inputs are ANSI according to the current code page; this is fine if the file paths can be converted to ANSI, but may fail to open some files and will use ANSI file paths inside .gltf files which violates glTF specification.

This can be worked around by using UTF8 as ANSI code page via a manifest; however, to avoid having to bundle a manifest we can also convert the command line arguments to UTF8 and set UTF8 as the codepage instead.

This requires Windows 10 1803; on earlier Windows versions setlocale should silently fail so this should be more or less equivalent to the previous code.

Fixes #767

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -58,6 +58,10 @@ set(GLTF_SOURCES
     gltf/write.cpp
 )
 
+if(WIN32)
+    list(APPEND GLTF_SOURCES gltf/gltfpack.manifest)
+endif()
+
 if(MSVC)
     add_compile_options(/W4)
 else()
--- a/gltf/gltfpack.cpp
+++ b/gltf/gltfpack.cpp
@@ -1217,8 +1217,12 @@ unsigned int textureMask(const char* arg)
 }
 
 #ifndef GLTFFUZZ
-int run(int argc, char** argv)
+int main(int argc, char** argv)
 {
+#ifndef __wasi__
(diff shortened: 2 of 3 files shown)


zeux:
This can also be solved with a manifest and a couple of changes in CMakeLists.txt it looks like; will test this tomorrow as an alternative.

PR #771, opened 2024-09-19
simplify: Enable border->locked and seam->locked collapses

zeux:
While locked vertices can't move by themselves, ideally we should be
able to move any vertex into a locked vertex; currently this is allowed
for manifold but not allowed for border/seam. This artificially limits
the allowed collapses which can lead to larger than necessary result or
worse than necessary error as the simplifier is forced to consider edges
with worse error.

To enable border collapses, we simply need to extend the picking logic;
border-border edges are symmetric so it is enough to validate them in
one direction, but border-locked edges may be forward or backward (as
in, they may align with the topological edge direction or be opposite to
it), so we need to check loop or loopback depending on which vertex is
border.

Seam->locked collapses require special handling to discover seam pair:
a seam-seam edge is bidirectional whereas a seam-locked edge is
unidirectional and may be connected to any wedge out of an arbitrary
count in the locked vertex. Because of this we can't use wedge[] to find
the seam pair anymore, and need to switch to using loop/loopback based
on whether the edge is forward or backward wrt original topology.

This is somewhat risky as it requires loop/loopback to stay valid and
coherent throughout the entire simplification process, whereas
previously bugs here would simply block certain collapses; now it's
theoretically possible to get an out of bounds read if the loop data for
the other wedge points nowhere.

To mitigate this, this code expands validation of this data with more
assertions and adds a fuzzer that tries to discover topology based bugs.
Additionally, there were two bugs in the classification of seam vertices and
in seam edge updates (the later bug by itself could artifically limit
collapses along seam loops in complex meshes) that have been fixed.

The effect of this change depends on the mesh. For smooth meshes with
few seams or borders the impact is minimal; for some meshes with a
reasonable balance of edge types this can result in significant improvements,
e.g. 20% fewer triangles at the same error or 15% smaller error at the same
triangle count for one of the meshes tested.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -216,6 +216,9 @@ codecbench-simd.wasm: tools/codecbench.cpp $(LIBRARY_SOURCES)
 codecfuzz: tools/codecfuzz.cpp src/vertexcodec.cpp src/indexcodec.cpp
 	$(CXX) $^ -fsanitize=fuzzer,address,undefined -O1 -g -o $@
 
+simplifyfuzz: tools/simplifyfuzz.cpp src/simplifier.cpp
+	$(CXX) $^ -fsanitize=fuzzer,address,undefined -O1 -g -o $@
+
 $(LIBRARY): $(LIBRARY_OBJECTS)
 	ar rcs $@ $^
 
--- a/demo/simplify.html
+++ b/demo/simplify.html
@@ -235,8 +235,8 @@
 								var a = dres[0][i + e],
 									b = dres[0][i + ((e + 1) % 3)];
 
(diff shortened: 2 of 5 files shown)


zeux:
As an example of a possible impact, here's before and after shots of a chair asset, simplified in auto-LOD mode with attribute weighting. These look basically identical, but "before" is 7680 triangles and "after" is 5514 triangles, at error parity.

image
image

When simplifying in manual mode to 10% of triangles, both are ~5600 triangles but "after" looks noticeably better.

image
image

This is all due to unlocking valuable collapses that are lower-error but were previously locked.

PR #772, opened 2024-09-25
Update USERS.md

corporateshark:
Added `LightweightVK` to the list of engines.


zeux:
Note that the list is currently limited to commercial or large open source projects (as documented at the top of the file); this is mostly so that it can aim to be comprehensive and not include very many smaller projects folks have. I'm not sure what the relationship is between LightweightVK and facebook/igl (of which it is a fork?), or whether facebook/igl uses meshoptimizer, but if it does maybe add that instead?

corporateshark:
Never mind.

Though, this project is featured in multiple Vulkanised 2024 presentations (at least two by Meta and at lease one by Khronos folks) and one SIGGRAPH 2024 talk. Vulkan Validation Layers are tested against this project.

PR #778, opened 2024-10-02
demo: Miscellaneous improvements to clustered simplification

zeux:
This change mainly reworks cluster boundary locking to use explicit lock flags and reworks DAG build a little bit to allow for customizable group size with more careful checks. In addition, partitioning code is now correctly taking into account vertices with the same position, which strengthens adjacency between neighboring clusters, and switches to vertex adjacency for simplicity.

The larger group size allows much more freedom, as METIS would more often generate at least partially connected groups with a larger size (I also wanted to investigate recursive partitioning instead of k-way partitioning but since the goal isn't really to maximize METIS quality this is probably good enough for now). Together with more precise boundary detection, we can now get cliffs.obj (2M triangles) all the way down to a single 3-cluster (383 triangle) group when using METIS=1 configuration:

Screenshot from 2024-10-01 20-23-45
Screenshot from 2024-10-01 20-23-50

Using METIS=2 is actually worse here, as although it generates smaller final result (255 triangles), the tree is much deeper (depth 26 instead of depth 16) and the clusters are ~90% filled on average. UE5 targets a slightly more conservative cluster size (usually I get ~126 triangles per cluster there) which I assume allows METIS a little bit more headroom to make mistakes with suboptimal cluster splits during a recursive bisection, something to try another time.

Note that `meshopt_simplifyWithAttributes` will currently only lock a vertex if *all* copies of the vertex with the same position are locked; if only some of them are locked and some aren't, locking isn't guaranteed to happen because only one of the vertices will be used for reference. The ideal interface contract is not fully clear to me here (eg in presence of `meshopt_SimplifySparse` flag we can't look at *all* copies), but it would probably be better if it was enough to lock any copy referenced by the index buffer... this change does not address this as it's easy enough to lock all copies on the boundary simultaneously.

### DIFF
--- a/demo/nanite.cpp
+++ b/demo/nanite.cpp
@@ -51,6 +51,7 @@ struct Cluster
 };
 
 const size_t kClusterSize = 128;
+const size_t kGroupSize = 8;
 const bool kUseLocks = true;
 
 static LODBounds bounds(const std::vector<Vertex>& vertices, const std::vector<unsigned int>& indices, float error)
@@ -335,7 +336,7 @@ static std::vector<std::vector<int> > partitionMetis(const std::vector<Cluster>&
 
 	int nvtxs = int(pending.size());
 	int ncon = 1;
-	int nparts = int(pending.size() + 3) / 4;
+	int nparts = int(pending.size() + kGroupSize - 1) / kGroupSize;
 	int edgecut = 0;
 
 	if (nparts <= 1)
@@ -375,7 +376,7 @@ static std::vector<std::vector<int> > partition(const std::vector<Cluster>& clus
 	// rough merge; while clusters are approximately spatially ordered, this should use a proper partitioning algorithm
 	for (size_t i = 0; i < pending.size(); ++i)
 	{
-		if (result.empty() || last_indices + clusters[pending[i]].indices.size() > kClusterSize * 4 * 3)
+		if (result.empty() || last_indices + clusters[pending[i]].indices.size() > kClusterSize * kGroupSize * 3)
 		{
 			result.push_back(std::vector<int>());
 			last_indices = 0;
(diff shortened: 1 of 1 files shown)


JMS55:
Awesome to see more progress here.
1. Larger group sizes - good to hear this helps. This has been on my list of things to try, as it's easy to tweak. Nice see it tested.
2. "Partitioning code is now correctly taking into account vertices with the same position" - can you explain this one a bit more to me? Do you mean that two vertices would have the same (~nearly the same?) position, but not share the same index (as in from an index buffer)? Why would this occur?
3. RE: Explicit locking rather than the automatic locking flag - another dev made a PR to my (Bevy's) codebase adding this https://github.com/JMS55/bevy/pull/28. When I tried it out, however, I found it very very aggressively collapsed edges, leading to few LOD levels (Ideally we want a perfect log2(n) distribution), with the final level always being a sphere or oval and not an actually good approximation of the mesh. I'll have to compare the code to yours, maybe it's bugged.

zeux:
> Do you mean that two vertices would have the same (~nearly the same?) position, but not share the same index (as in from an index buffer)? Why would this occur?

This can happen if an attribute seam happens to lie on the cluster boundary (for example, a UV seam where there are two vertices with two different indices, the same position, but two different UV coordinates).

zeux:
As for the boundary, that code doesn't seem to reset lock flags to `false` on any edges other than the mesh boundary; my code computes the boundary between all groups at a given level at once from scratch, but it's also possible to reset the flags back to false. Here's cliffs.obj with kUseLocks=false (2371 triangles on smallest LOD) vs kUseLocks=true (383 triangles on smallest LOD).

Screenshot From 2024-10-02 19-27-41

Screenshot From 2024-10-02 19-27-55

SparkyPotato:
> As for the boundary, that code doesn't seem to reset lock flags to false on any edges other than the mesh boundary; this would permanently keep all interior boundaries locked, including the original cluster boundaries, unless I am misreading it, which would quickly terminate simplification as no edges will be collapsible

Ah, `group_boundary` is never reset back to `false`, good catch! That should fix the issues in simplification.

PR #780, opened 2024-10-03
vertexcodec: Reintroduce tracing and add roundtrip fuzzing

zeux:
This change reintroduces TRACE macro to vertex encoder; we used to have tracing code but it was removed in 2020 as the v0 codec was finalized. However, in order to make progress towards v1 codec we need to bring some of it back; this change adds byte tracking back with a more concise but complete summary display, and also adds bit consistency tracking (a bit is considered consistent if it is the same as the last bit in the same position for the entire byte group). Bit consistency tracking helps analyze opportunities where rotation or reorganization of data can improve compression.

Note that tracing relies on global state and is not thread-safe; since it prints data to stdout it's not really useful in a threaded context either way, and the thread safety issues should not cause any runtime issues, just incorrect data aggregation (nor will tracing ever be enabled by default or exposed as an official configuration, so it's mostly moot).

Also add roundtrip fuzzing to the existing codec fuzzer; if we were to implement v1, we would need a way to ensure data is compressed losslessly.

### DIFF
--- a/src/vertexcodec.cpp
+++ b/src/vertexcodec.cpp
@@ -1210,7 +1210,7 @@ size_t meshopt_encodeVertexBuffer(unsigned char* buffer, size_t buffer_size, con
 	{
 		const Stats& vsk = vertexstats[k];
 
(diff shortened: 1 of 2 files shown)


zeux:
Example trace, 24 byte vertex, 12 byte float3 position, two ushort2 UV channels, packed normal (1-11-10-10):

image

PR #786, opened 2024-10-07
demo: Improve boundsError to take camera parameters into account

zeux:
While boundsError is an approximation and the precise version would compute the projected sphere bounds taking into account perspective distortion and near plane clipping, our simpler variant can work as well and is invariant to camera rotation which is a nice property in certain cases; for example in VR you might want to compute LOD based on a "third" eye, a center camera, so that the geometry is in sync between two eyes.

Our boundsError implementation was mostly serviceable but did not depend on camera field of view and had some potential issues with monotonicity for cases where the camera position was inside the sphere.

We now clip the distance to znear and properly scale it so that the result is a normalized screen space error (0..1, multiply by screen height to get pixels). This ensures that if the bounds.error is 0, boundsError will return 0 even if the camera is inside, which should ensure proper cluster selection for extreme closeups.

### DIFF
--- a/demo/main.cpp
+++ b/demo/main.cpp
@@ -197,7 +197,7 @@ void dumpObj(const std::vector<Vertex>& vertices, const std::vector<unsigned int
 	{
 		unsigned int a = indices[i], b = indices[i + 1], c = indices[i + 2];
 
-		fprintf(stderr, "f %d %d %d\n", a + 1, b + 1, c + 1);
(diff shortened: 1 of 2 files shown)


zeux:
Preliminary normal support; this is with a fairly low normal weight and ~2px error bound so shading issues remain. Note that the LOD selection here is only performed every ~10% of the video, because this was rendered in Blender by manually keyframing camera motion and cuts visibility :sweat_smile:

[0001-0110.webm](https://github.com/user-attachments/assets/edea55b3-22b6-41b4-923f-8aed450a2386)

PR #794, opened 2024-10-22
clusterizer: Refactor and improve triangle prioritization

zeux:
When getNeighborTriangle's best selection can't fit into the current
meshlet, we redo the search process using an adjusted criteria that
ignores position/normal as that is irrelevant. However, internally
getNeighborTriangle adjusted best_extra to act as a topology score
instead of just the extra count. This would lead to the last triangle of
the meshlet being reselected in addition to the first one in some cases.

This led to confusing logic but also accidentally uncovered an opportunity
for improvement, as rescoring could penalize local selection but overall
resulted in better clusterization due to better global flow. We attempt to
capitalize on that discovery further by adding a rule for valence=2 edges,
as prioritizing these triangles seems to improve clusterization further
(1-2% fewer meshlets in some cases with a little bit better packing, reduced
boundary edges, and 1.5x-2x fewer disconnected clusters).


zeux:
This doesn't seem to affect Nanite strongly either way (the reduction in disconnected clusters mostly affects the first level and it's still important to group clusters aggressively, plus METIS still requires a large group size to not produce disconnected groups), but the behavior is generally better for non-Nanite pipelines. As a somewhat extreme example, cliffs.obj goes from 21600 meshlets to 21112 (-2%), with 1.5x fewer disconnected clusters when using 64/124 setup (non-Nanite).

zeux:
master
```
Meshlets : 309 meshlets (avg vertices 63.8, avg triangles 93.7, avg boundary 29.3, not full 4, not connected 25) in 8.95 msec
Meshlets : 357 meshlets (avg vertices 63.9, avg triangles 88.4, avg boundary 26.3, not full 8, not connected 87) in 8.76 msec
Meshlets : 386 meshlets (avg vertices 63.9, avg triangles 88.2, avg boundary 25.5, not full 8, not connected 107) in 9.31 msec
Meshlets : 738 meshlets (avg vertices 63.9, avg triangles 94.3, avg boundary 31.3, not full 2, not connected 47) in 22.81 msec
```

this PR at the time of this comment
```
Meshlets : 306 meshlets (avg vertices 63.8, avg triangles 94.6, avg boundary 28.6, not full 17, not connected 10) in 9.37 msec
Meshlets : 350 meshlets (avg vertices 63.9, avg triangles 90.2, avg boundary 24.5, not full 22, not connected 75) in 8.87 msec
Meshlets : 379 meshlets (avg vertices 63.8, avg triangles 89.9, avg boundary 23.7, not full 24, not connected 94) in 9.54 msec
Meshlets : 731 meshlets (avg vertices 63.9, avg triangles 95.3, avg boundary 30.5, not full 20, not connected 31) in 23.69 msec
```

also exploring some further improvements to heuristics, intermediate results (WIP, not conclusive yet)
```
Meshlets : 304 meshlets (avg vertices 63.8, avg triangles 95.2, avg boundary 27.9, not full 18, not connected 10) in 9.60 msec
Meshlets : 347 meshlets (avg vertices 63.8, avg triangles 90.9, avg boundary 23.5, not full 32, not connected 72) in 10.13 msec
Meshlets : 376 meshlets (avg vertices 63.9, avg triangles 90.6, avg boundary 23.0, not full 39, not connected 88) in 12.16 msec
Meshlets : 723 meshlets (avg vertices 63.9, avg triangles 96.3, avg boundary 29.5, not full 42, not connected 20) in 34.01 msec
```

zeux:
All of the above are a little aggressive wrt leaving the last vertex of the meshlet unused, but there's some tweaks that are less significantly reducing disconnected clusters at the cost of a better fill, e.g. `priority = 1 + extra` in the newly added case yields this which is still better than master on every metric:

```
Meshlets : 306 meshlets (avg vertices 63.8, avg triangles 94.6, avg boundary 28.5, not full 5, not connected 18) in 9.93 msec
Meshlets : 352 meshlets (avg vertices 63.9, avg triangles 89.6, avg boundary 25.1, not full 7, not connected 83) in 9.58 msec
Meshlets : 380 meshlets (avg vertices 63.9, avg triangles 89.6, avg boundary 24.1, not full 8, not connected 108) in 10.22 msec
Meshlets : 728 meshlets (avg vertices 64.0, avg triangles 95.6, avg boundary 30.3, not full 1, not connected 28) in 25.11 msec
```

zeux:
Checking a few of these meshes on niagara, and `1 + extra` variant improves end to end rendering performance, whereas `2` regresses it a little bit so will go with the former as a safer version for now. More work maybe pending in future PRs.

JMS55:
You may be interested in this other implementation of virtual geometry I was told about recently. They use some interesting techniques for grouping triangles into clusters and clusters into groups with METIS (page ~20).

https://pettett.github.io/projects/meshes/main.pdf
https://github.com/pettett/multires/tree/main/baker/src

zeux:
FWIW this PR is unrelated to Nanite for the most part (its main purpose is to improve clustering in general, not specifically for Nanite). I wasn't aware of that thesis; for Nanite demo, my plans do not involve METIS and it looks like they use METIS with some more involved tweaks, but I guess it may be valuable to incorporate them if they are fairly straightforward to have a better baseline.

zeux:
After further experiments with a priority table solver it looks like we're close to exhausting reasonable heuristics, at least in the limited local configuration space. For any individual mesh, a solver can find a more optimal configuration (e.g. 1% further reduction in meshlet count and further reduction in disconnected clusters), but these tweaks don't seem to generalize very well. Perhaps a more thorough search can help, e.g. I only tried looking for rule sets using number of low-valence vertices in the triangles but perhaps there's more fine grained analysis that could help, like distinguishing valence-1 vertices that are used vs not-used in the current meshlet. Regardless, I'm happy with the current state within the confines of a greedy walk for now -- further improvements in clusterization probably require a non-greedy algorithm.

PR #795, opened 2024-10-23
Release 0.22

zeux:
- Updates all version numbers to 0.22
- Stabilizes decode/encode filters with the exception of recently added meshopt_EncodeExpClamped

For posterity, APIs that are still experimental:

- `meshopt_generateProvokingIndexBuffer`: added in this release, unclear if it will need changes
- `meshopt_EncodeExpClamped` enum: added in this release, unclear if it will stay (cutoff fixed to 0) or be replaced by `meshopt_encodeFilterExpClamped` function (customizable cutoff)
- `meshopt_SimplifyPrune` enum: added in this release, implementation is still in flux with likely behavior changes in future releases
- `meshopt_simplifyWithAttributes`: a lot of implementation changes in this release; while the implementation should be relatively stable, it's unclear if interface changes will be needed for at least one planned feature
- `meshopt_simplifySloppy`: hasn't changed in a few releases but significant implementation work may be outstanding
- `meshopt_simplifyPoints`: a few implementation changes in this release; likely effectively stable, unless we decide we need to support RGB/RGBA modes which would need an interface change to specify if the input color is 3-float or 4-float
- `meshopt_optimizeMeshlet`: no changes since last release, but would like to wait for one more release to stabilize in case clusterizer sees significant changes in the next release.
- `meshopt_spatialSortTriangles`: no changes in a while; `meshopt_spatialSortRemap` has been stabilized and is useful for point clouds, but there's some issues with sorting triangles for compression efficiency that I'd like to look into before stabilizing this, as right now it's unclear if the triangle sort is practically useful.

As a reminder, "experimental" means the interface or implementation may see significant change; non-experimental means interface is de-facto locked down (both API/ABI) and implementation should be production ready, but future releases may still improve behavior/performance in a compatible manner. Experimental APIs are still fine to use, it's just that the compatibility isn't guaranteed.

`meshopt_generateProvokingIndexBuffer`, `meshopt_simplifyWithAttributes`, `meshopt_simplifyPoints` and `meshopt_optimizeMeshlet` will most likely become stable in the next release.


zeux:
After doing some more pre-release testing and review I think we're set for a release tomorrow morning.

PR #796, opened 2024-10-24
js: Fix vertex_lock support for simplifyWithAttributes

zeux:
This code never worked if an actual array was passed in as it required an element per float instead of an element per vertex.

Since we need to change this anyway, change the interface to accept Uint8Array instead of a boolean array. This is more consistent with other parameters and allows an upgrade path if the C interface changes from "lock flags" to just "flags".


zeux:
Also somewhat tempted to move vertex_lock to the end of the argument list for JS to make this API more idiomatic... not sure if this is worth diverging from C++ though.

PR #798, opened 2024-10-28
gltfpack: Allow per-texture-class control for texture scaling

zeux:
While we expose an ability to specify texture compression settings for each texture class separately, there's a significant quality and size gap between ETC1S and UASTC that doesn't have any middle-ground options.

In some cases, downscaling an UASTC texture looks much better than a full-resolution ETC1S; however, when ETC1S provides sufficient quality, this can in turn reduce the quality visibly.

This change makes it possible to adjust the tradeoff with more granularity by specifying texture scaling per-class; for example,

	gltfpack -tu -tc color -ts attrib 0.75

... will use ETC1S for color textures at full resolution, UASTC for other textures (normal maps and attribute textures), but will further downscale attribute textures (roughness, etc.) to 0.75 x 0.75 of the original resolution (which is around 2x smaller).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -100,7 +100,7 @@ jobs:
           curl -sL $VALIDATOR | tar xJ
           find glTF-Sample-Assets -name '*.gltfpack.gltf' | xargs -P 2 -L 1 -d '\n' ./gltf_validator -r -a
         env:
-          VALIDATOR: https://github.com/KhronosGroup/glTF-Validator/releases/download/2.0.0-dev.3.8/gltf_validator-2.0.0-dev.3.8-linux64.tar.xz
+          VALIDATOR: https://github.com/KhronosGroup/glTF-Validator/releases/download/2.0.0-dev.3.10/gltf_validator-2.0.0-dev.3.10-linux64.tar.xz
 
   gltfpack-js:
     runs-on: ubuntu-latest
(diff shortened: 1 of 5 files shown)


zeux:
Motivating screenshots using FlightHelmet from glTF-Asset-Models (sizes and screenshots courtesy of [gltf.report](https://gltf.report/)) below. I consider full ETC1S option not super useful; ETC1S-UASTC hybrid is great but doesn't reduce the disk size much. The only alternative that existed before this change would be to downsample everything (last image), but now there's a host of healthy options to strike pretty much an arbitrary size:quality balance.

(note, I'm ignoring `-tq` setting for UASTC here, as it doesn't change VRAM consumption, but there's also an option of downgrading UASTC quality per channel to reduce disk size that exists today)

- original textures (47 MB disk, 286 MB vram)

image

- ETC1S `-tc` (8 MB disk, 38 MB vram)

image

- UASTC `-tu` (41 MB disk, 73 MB vram)

image

- ETC1S color, UASTC everything else `-tu -tc color` (32 MB disk, 62 MB vram)

image

- ETC1S color, UASTC everything else but 2x smaller `-tu -tc color -ts normal,attrib 0.5` (12 MB disk, 26 MB vram)

image

- ETC1S color, UASTC everything else, normal maps 0.75x, attrib 0.5x `-tu -tc color -ts normal 0.75 -ts attrib 0.5` (16 MB disk, 33 MB vram)

image

- ETC1S color, UASTC everything else, everything downscaled 2x `-tu -tc color -ts 0.5` (10 MB disk, 16 MB vram)

image

arpu:
this is  very helpful, one question if i use -tl like -tl 1024 max image resolution 1k  the -ts normal 0.5 will scale to 1024/2 or to the input resolution?

zeux:
The limit would be applied after the scaling (same behavior as it is right now)

arpu:
would be nice, to scale the normal, in this example to 512x512

zeux:
Yeah this would require per class limits as well. Which is a good idea, will do as a separate PR (eg `-tl 1024 -tl normal 512`)

PR #802, opened 2024-11-13
gltfpack: Take account of primitive extras

### LINKED ISSUE #808: gltfpack: Support primitive extras when -ke is specified
GLTF format specifications allows to add "extras" in primitives, gltfpack ignore it and merge primitives even when extras are present and different.

JulienIcon:
GLTF format allows extras in primitives, gltfpack ignore it and merge primitives even when extras are present and different.

Fix #808

### DIFF
--- a/gltf/mesh.cpp
+++ b/gltf/mesh.cpp
@@ -238,7 +238,7 @@ static bool canMergeMeshes(const Mesh& lhs, const Mesh& rhs, const Settings& set
 	if (lhs.streams.size() != rhs.streams.size())
 		return false;
 
-	if (lhs.extras.data && rhs.extras.data && strcmp(lhs.extras.data, rhs.extras.data))
+	if (settings.keep_extras && !areExtrasEqual(lhs.extras, rhs.extras))
 		return false;
 
 	for (size_t i = 0; i < lhs.streams.size(); ++i)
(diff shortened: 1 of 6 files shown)


gltf/mesh.cpp
```diff
@@ -238,6 +238,9 @@ static bool canMergeMeshes(const Mesh& lhs, const Mesh& rhs, const Settings& set
 	if (lhs.streams.size() != rhs.streams.size())
 		return false;
 
+	if (lhs.extras.data && rhs.extras.data && strcmp(lhs.extras.data, rhs.extras.data))
```
zeux, line 241: This comparison should only work if keep_extras is set. Also, this will not work if one of the mesh has empty extras. See material.cpp/areExtrasEqual for the correct way to implement this.

zeux:
Also missing some support in deduplication, probably enough to disable deduplication by returning `false` from `canDedupMesh` - again, only if `keep_extras` is set.

JulienIcon:
I take account of your comments but I also added :

- writing, only if keep_extras is set, the scene extras
- writing the mesh name

zeux:
I'm not interested in merging mesh name preservation, as meshes are very ephemeral and they get reconstructed by many different optimizations in gltfpack.

Scene extras are fine I guess. I don't see why they should be part of this PR, but it's just a couple of lines so maybe it's fine.

PR #810, opened 2024-12-04
vertexcodec: Initial implementation of experimental next version

zeux:
This PR prototypes initial support for the next vertex codec version.

The change introduces version 0xe; this version has zero compatibility guarantees, and will be in the state of flux wrt bitstream format until it is finalized as version 1. The goals for the next version are to retain comparable decoding performance to v0 -- currently v0 decodes around 4.5 GB/s on Zen 4 (7950X), around 3.5 GB/s on M2 (base) -- and to increase compression ratio, targeting up to ~10% relative improvement on average for a set of geometry data. This will make the decoder more complicated and larger, but hopefully not dramatically so. Some performance loss is probably unavoidable, hopefully it can be kept at ~5% or so.

In this PR the initial infrastructure (version selection, enhanced codectest harness) is setup, and byte group encoding is expanded to support a palette of bit counts, adding support for 1-bit and 6-bit groups as well as literal byte blocks. This results in aggregate improvements as follows on the test set (~half of this is glTF data, ~half is proprietary engine data).

> 143 files: raw v1/v0 -4.83%, lz4 v1/v0 -3.37%, zstd v1/v0 -0.79%

~It's not certain yet that the decoding performance can be maintained with this expansion due to some issues around codegen with compilers, but that will be investigated separately. To get to the target compression ratio improvements, more changes will be necessary but they will be orthogonal to byte group encoding enhancements which this PR will focus on.~

Unfortunately, introduction of 6-bit groups significantly penalized decoding performance across the board; while they are useful on some data, to maintain excellent performance the approach had to be adjusted: only 1-bit groups are added for byte group encoding, but control bits are enhanced to support zero channels as well as literal channels within a block. This as well as rebalanced dispatch structure allows to maintain excellent decoding performance, but results in slightly less significant ratio wins unless zstd is used:

> 143 files: raw v1/v0 -3.70%, lz4 v1/v0 -2.43%, zstd v1/v0 -0.81%

... that said, decoding performance is actually often *better*, and on just game ready engine data the wins are still decent, and actually better post-zstd:

> 64 files: raw v1/v0 -4.06%, lz4 v1/v0 -2.99%, zstd v1/v0 -1.12%

Further tweaks may be possible to get more performance out of the new dispatch structure, especially for `gcc` which leaves a bounds check for switch table even though it's provably unnecessary.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -88,6 +88,10 @@ ifeq ($(config),coverage)
 	LDFLAGS+=-coverage
 endif
 
+ifeq ($(config),release-avx512)
+	CXXFLAGS+=-O3 -DNDEBUG -mavx512vl -mavx512vbmi -mavx512vbmi2
+endif
+
 ifeq ($(config),release-scalar)
 	CXXFLAGS+=-O3 -DNDEBUG -DMESHOPTIMIZER_NO_SIMD
 endif
--- a/demo/main.cpp
+++ b/demo/main.cpp
@@ -847,12 +847,56 @@ void encodeVertex(const Mesh& mesh, const char* pvn, bool validate = true)
 
 	size_t csize = compress(vbuf);
 
-	printf("VtxCodec%1s: %.1f bits/vertex (post-deflate %.1f bits/vertex); encode %.2f msec, decode %.2f msec (%.2f GB/s)\n", pvn,
+	printf("VtxCodec%1s%s: %.1f bits/vertex (post-deflate %.1f bits/vertex); encode %.2f msec, decode %.2f msec (%.2f GB/s)\n", pvn,
+	    res == 0 && memcmp(&pv[0], &result[0], pv.size() * sizeof(PV)) == 0 ? "" : "!",
 	    double(vbuf.size() * 8) / double(mesh.vertices.size()),
 	    double(csize * 8) / double(mesh.vertices.size()),
 	    (middle - start) * 1000,
 	    (end - middle) * 1000,
 	    (double(result.size() * sizeof(PV)) / (1 << 30)) / (end - middle));
+}
+
+template <typename PV>
+void benchmarkVertex(const Mesh& mesh, const char* pvn)
+{
+	std::vector<PV> pv(mesh.vertices.size());
+	packMesh(pv, mesh.vertices);
+
+	// allocate result outside of the timing loop to exclude memset() from decode timing
+	std::vector<PV> result(mesh.vertices.size());
+
+	std::vector<unsigned char> vbuf(meshopt_encodeVertexBufferBound(mesh.vertices.size(), sizeof(PV)));
+	vbuf.resize(meshopt_encodeVertexBuffer(&vbuf[0], vbuf.size(), &pv[0], mesh.vertices.size(), sizeof(PV)));
+
+	// warmup
+	meshopt_decodeVertexBuffer(&result[0], mesh.vertices.size(), sizeof(PV), &vbuf[0], vbuf.size());
+
+	for (int run = 0; run < 10; ++run)
+	{
+		double mint = 0;
+		double avgt = 0;
+		double vart = 0;
+
+		for (int i = 0; i < 100; ++i)
+		{
+			double start = timestamp();
+			meshopt_decodeVertexBuffer(&result[0], mesh.vertices.size(), sizeof(PV), &vbuf[0], vbuf.size());
+			double end = timestamp();
+
+			double time = end - start;
+
+			mint = (mint == 0 || time < mint) ? time : mint;
+
+			// Welford variance computation
+			double delta = time - avgt;
+			avgt += delta / (i + 1);
+			vart += delta * (time - avgt);
+		}
+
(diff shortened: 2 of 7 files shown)


zeux:
Subsequent commits in this PR introduce decoding implementation which covers scalar and all SIMD targets; it has had a limited amount of optimizations, mostly focusing on clang/SSSE3 and clang/NEON.

Some compilers hit significant codegen issues with the new code, e.g. gcc/Zen4 is inflated because of some inlining issues I believe. The switch dispatch got more expensive, which is partially mitigated by the new UNREACHABLE macro (previously compilers deduced this fact from the limited selector bits).

Unfortunately, performance penalty remains severe: since the ratio only improves by 5%, it would be optimal to have ~5% performance loss here, but no compiler on no architecture reaches that. A big contributor to this is bit 6 decoding mode: it often replaces bit 8 in an attempt to squeeze a few more bytes out of the encoding, but is much much much more expensive to decode (it's the most expensive bit mode now, as 6 bits can't be cleanly packed into bytes).

This can be partially mitigated by removing sentinel support from bit6, but that reduces efficiency gains. Removing bit6 mode outright reduces efficiency gains even further...

This is a developing space; more commits will hopefully follow as I get new ideas.

| Decode            | v0 clang | v1 clang    | v0 gcc | v1 gcc      | v0 msvc | v1 msvc     |
| ----------------- | ------- | ----------- | ----- | ----------- | ------ | ----------- |
| SSSE3 (Zen4)      | 5.51    | 4.63 (-15%) | 5.52  | 3.76 (-32%) | 5.29   | 3.95 (-25%) |
| AVX512 (Zen4)     | 6.45    | 5.65 (-14%) | 6.11  | 4.80 (-21%) | 5.04   | 4.84 (-4%)  |
| SSSE3 (Ice Lake)  | 2.68    | 2.39 (-11%) | 2.36  | 2.17 (-8%)  | -      | -           |
| AVX512 (Ice Lake) | 2.87    | 2.64 (-8%)  | 2.84  | 2.52 (-13%) | -      | -           |
| NEON (M2)         | 3.46    | 3.02 (-13%) | -     | -           | -      | -           |
| NEON (Graviton3)  | 2.62    | 2.14 (-19%) | 2.35  | 2.03 (-14%) | -      | -           |
| WASM (Zen4)       | 2.82    | 2.25 (-20%) | -     | -           | -      | -           |

zeux:
Okay I think I know where to take all of this. This PR will be closed later, but I'll pull some commits from this into a new PR series which will approach things a little differently and with better results.

zeux:
Alright, I've had to revise this pretty substantially, as decoding 6-bit groups runs into a lot of performance challenges that make the whole algorithm too much slower. 15% degradation is too much, as this is only the first change, and 6-bit in particular are problematic (details outside of scope of this PR).

With 6-bit groups gone, we have one palette entry left; it's now used for a zero run, and the palette is restructured to either be 0/1/2/4 or 1/2/4/8 bits in v1 (still 0/2/4/8 bits in v0 obviously). Supporting this without adding dispatch overhead is basically impossible but I think I've managed to minimize that during decoding. As an extra bonus, v1 can actually decode faster now, as literal and zero blocks can run specialized paths through memset/memcpy.

The size gains are less significant unfortunately. On the full data set I get:

> Before: 143 files: raw v1/v0 -4.83%, lz4 v1/v0 -3.37%, zstd v1/v0 -0.79%
> After: 143 files: raw v1/v0 -3.70%, lz4 v1/v0 -2.43%, zstd v1/v0 -0.81%

However, post-zstd sizes are the same, and on runtime engine data I get:

> Before: 64 files: raw v1/v0 -4.53%, lz4 v1/v0 -3.31%, zstd v1/v0 -0.88%
> After: 64 files: raw v1/v0 -4.06%, lz4 v1/v0 -2.99%, zstd v1/v0 -1.12%

Which is a less severe drop, and actually things are better post-zstd.

The tradeoff is that the decoding performance penalty is much smaller now. With integrated v0 & v1 decoders through switch based dispatch there's somewhere between 0 and 5% overhead on decoding v0 streams, but v1 can actually pull ahead.

For example, with clang++ / SSSE3 / cliffs.obj test asset in demo program, I get:

With 6-bit decoding:
v0 decodes at 4.70 GB/s
v1 decodes at 3.56 GB/s

With new decoding:
v0 decodes at 4.60 GB/s
v1 decodes at 5.00 GB/s

gcc in particular runs into some codegen issues that are shared between the two approaches, but gcc is not a production compiler from the performance perspective and I suspect these can be fixed as it looks like a bug in the optimizer. NEON and Wasm SIMD see no performance penalty for v0 in this update from what I can tell when using clang either. MSVC sees a small (<5%) penalty in v0 which is recovered by v1.

So overall 6-bit groups are just too problematic despite the size gains, thus they are out. This PR should be good to merge as for now we'll assume this is the decoding structure going forward. In the future maybe the switch based dispatch can be a little better but that doesn't require format changes.

zeux:
Further tweaks may be possible to get more performance out of the new dispatch structure, especially for `gcc` which leaves a bounds check for switch table even though it's provably unnecessary:

image

but that can wait until future PRs.

PR #811, opened 2024-12-06
vertexcodec: Initial implementation of experimental decoders

zeux:
This PR prototypes initial support for the next vertex codec version.

The change introduces version 0xe; this version has zero compatibility guarantees, and will be in the state of flux wrt bitstream format until it is finalized as version 1. The goals for the next version are to retain comparable decoding performance to v0 -- currently v0 decodes around 4.5 GB/s on Zen 4 (7950X), around 3.5 GB/s on M2 (base) -- and to increase compression ratio, targeting up to ~10% relative improvement on average for a set of geometry data. This will make the decoder more complicated and larger, but hopefully not dramatically so. Some performance loss is probably unavoidable, hopefully it can be kept at ~5% or so.

In this PR the initial infrastructure (version selection, enhanced codectest harness) is setup, and byte group encoding is expanded to support a palette of bit counts, adding support for 1-bit and 6-bit groups as well as literal byte blocks. This results in aggregate improvements as follows on the test set (~half of this is glTF data, ~half is proprietary engine data).

    143 files: raw v1/v0 -4.83%, lz4 v1/v0 -3.37%, zstd v1/v0 -0.79%

It's not certain yet that the decoding performance can be maintained with this expansion due to some issues around codegen with compilers, but that will be investigated separately. To get to the target compression ratio improvements, more changes will be necessary but they will be orthogonal to byte group encoding enhancements.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -88,6 +88,14 @@ ifeq ($(config),coverage)
 	LDFLAGS+=-coverage
 endif
 
+ifeq ($(config),release-avx)
+	CXXFLAGS+=-O3 -DNDEBUG -mavx
+endif
+
+ifeq ($(config),release-avx512)
+	CXXFLAGS+=-O3 -DNDEBUG -mavx512vl -mavx512vbmi -mavx512vbmi2
+endif
+
 ifeq ($(config),release-scalar)
 	CXXFLAGS+=-O3 -DNDEBUG -DMESHOPTIMIZER_NO_SIMD
 endif
--- a/demo/main.cpp
+++ b/demo/main.cpp
@@ -847,12 +847,53 @@ void encodeVertex(const Mesh& mesh, const char* pvn, bool validate = true)
 
 	size_t csize = compress(vbuf);
 
-	printf("VtxCodec%1s: %.1f bits/vertex (post-deflate %.1f bits/vertex); encode %.2f msec, decode %.2f msec (%.2f GB/s)\n", pvn,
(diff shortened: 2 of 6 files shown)


zeux:
| Decode            | v0clang | v1 clang    | v0gcc | v1 gcc      | v0msvc | v1 msvc     |
| ----------------- | ------- | ----------- | ----- | ----------- | ------ | ----------- |
| SSSE3 (Zen4)      | 5.51    | 4.63 (-15%) | 5.52  | 3.76 (-32%) | 5.29   | 3.95 (-25%) |
| AVX512 (Zen4)     | 6.45    | 5.65 (-14%) | 6.11  | 4.80 (-21%) | 5.04   | 4.84 (-4%)  |
| SSSE3 (Ice Lake)  | 2.68    | 2.39 (-11%) | 2.36  | 2.17 (-8%)  | -      | -           |
| AVX512 (Ice Lake) | 2.87    | 2.64 (-8%)  | 2.84  | 2.52 (-13%) | -      | -           |
| NEON (M2)         | 3.46    | 3.02 (-13%) | -     | -           | -      | -           |
| NEON (Graviton3)  | 2.62    | 2.14 (-19%) | 2.35  | 2.03 (-14%) | -      | -           |
| WASM (Zen4)       | 2.82    | 2.25 (-20%) | -     | -           | -      | -           |

zeux:
Actually to simply this I will update the other PR instead and close this one; we don't need two PRs. I was originally hoping I can submit encoding code separately, but the decoding performance situation makes the form of the new version uncertain.

PR #819, opened 2024-12-10
gltfpack: Disable animation resampling when frequency is 0

### LINKED ISSUE #815: gltfpack: Animation without resampling
Hi,

We use gltfpack to reduce the number of meshes but we have an unexpected sampling of animation (we use cubic spline). Is it possible to have an option to let untouched the animation without any resampling ?

Best regards

zeux:
Using `-af 0` now disables resampling and keeps the original time values.

When resampling is disabled, we can still mark tracks as constant and
remove them if they are not necessary, which cuts down on the animation
size. In this mode it is more difficult to support ranges, so we assume track
range will cover the relevant timespan if any tracks with time survive.

Doing this analysis for cubic splines is slightly more difficult as if
tangents are non-zero, an otherwise "constant" track will not result in
constant output, so we disable analysis in that case for simplicity.

Fixes #815.

### DIFF
--- a/gltf/animation.cpp
+++ b/gltf/animation.cpp
@@ -308,6 +308,7 @@ void processAnimation(Animation& animation, const Settings& settings)
 
 		track.time.clear();
 		track.data.swap(result);
+		track.interpolation = track.interpolation == cgltf_interpolation_type_cubic_spline ? cgltf_interpolation_type_linear : track.interpolation;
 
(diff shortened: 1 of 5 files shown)


zeux:
Hmm, range preservation can be reimplemented more cleanly now as it can be done during processAnimation. This would make it work even if resampling is turned off. Will do this before merging this.

PR #823, opened 2024-12-18
vertexcodec: Augment XOR delta encoding with rotation

zeux:
This can be useful to align the bits better in certain cases. However,
rotates come with a set of questions: what's the granularity? which
delta modes do they work in? how much do they cost to decode?

Ideally perhaps we would do fully orthogonal rotates. However, rotates
can be done as a filter step; as such, from the decoder perspective it's
best to focus on rotate applications that are minimal.

When combined with SUB deltas, rotates must be done after deltas get
recombined. This makes them expensive, as this step is ~scalarized. When
using XOR deltas, rotates can be done before or after undelta; that
makes them cheaper and aligns XOR decoding with other delta forms, as
the cost is similar to unzigzag.

Conceptually, there's also some value in this split: SUB deltas assume
integer-like values and bit propagation from MSB to LSB, so bit
alignment is expected; for arbitrary bitpacked data, SUB deltas end up
crossing packing thresholds so they are not optimal.

With XOR deltas, we could still choose to encode 8-bit or 16-bit
rotates. However, since SIMD ISAs commonly lack per-lane rotates, 16-bit
rotates require fixing both halves to do the same rotation which is not
really better than a 32-bit rotate; 8-bit rotates could be done before
transposition but require 12 bits (3x4) of extra channel encoding, 4x more
time to encode, are not much stronger vs a single 32-bit rotate that neatly
fits into existing channel encoding.

This change makes encoding substantially slower; however, future PRs are
expected to improve this as well as introduce support for compression
levels.

### DIFF
--- a/demo/main.cpp
+++ b/demo/main.cpp
@@ -824,7 +824,7 @@ void packVertex(const Mesh& mesh, const char* pvn)
 }
 
 template <typename PV>
-void encodeVertex(const Mesh& mesh, const char* pvn, bool validate = true)
+void encodeVertex(const Mesh& mesh, const char* pvn)
 {
 	std::vector<PV> pv(mesh.vertices.size());
 	packMesh(pv, mesh.vertices);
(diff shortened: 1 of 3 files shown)


zeux:
There's some uncertainties with whether or not rotation will make it into final v1 format; this depends on faster encoding support and further analysis on decoding performance tradeoffs between having rotates and not having rotates and using XOR for fast-decode paths. However in the interest of streamlining the development I'm going to merge this speculatively.

If further analysis will determine that rotation doesn't sufficiently pull its weight then it will be removed. Some other commits in this PR would be nice to have regardless so it's probably reasonable to do this in a separate commit when v1 gets finalized.

PR #829, opened 2025-01-03
vertexcodec: Additional latency optimizations for delta decoding

zeux:
The second half of the delta decoding loop relies on 16 serialized instructions that
incrementally compute 4 prefix sums and store the results. On Apple CPUs, this is a critical
path of the loop: subsequent iterations depend on updated value of `pi`, and this restricts
throughput.

While this could be worked around by doing parallel prefix sum computations (for each
of the 4 channels individually), that only really works for byte groups. A partial parallel
prefix sum that computes prefix sums in groups of 4 values and then adjusts them with
serial accumulation was implemented in the first commit in this PR; that makes full decoding
~20% faster on Apple M2.

Unfortunately, that approach does not translate to any other CPU - it's slower on Zen 3 and
Graviton 2 - and requires a completely separate implementation. Thus, this change switches
to a different method in a later commit, which recomputes `pi` from individual elements - this
is easier to do than a full prefix sum, since we only need to perform 4 parallel reductions (for
bytes) and that can be done fairly efficiently.

On Apple M2, this is even faster than the original method, resulting in up to 25% faster decoding,
`codecbench` reports 3.85 GB/s => 4.89 GB/s. The impact on just delta decoding is much stronger,
`decodeDeltas4Simd` gets ~1.7x faster.

This change does not translate to wins on other CPUs either (~2% regression on Zen 3, ~3%
regression on Graviton 2), but it is easier to maintain alongside other code. So for now we only
enable this on Apple CPUs to ensure this is a net positive.

(also a more minor optimization was committed separately in https://github.com/zeux/meshoptimizer/commit/abcfe76a0cd8dbeb2bc94c59bcb9ba8ac526ff58)

### DIFF
--- a/src/vertexcodec.cpp
+++ b/src/vertexcodec.cpp
@@ -1349,6 +1349,21 @@ inline uint8x16_t rotate32(uint8x16_t v, int r)
 	uint32x4_t v32 = vreinterpretq_u32_u8(v);
 	return vreinterpretq_u8_u32(vorrq_u32(vshlq_u32(v32, vdupq_n_s32(r)), vshlq_u32(v32, vdupq_n_s32(r - 32))));
 }
+
+template <int Channel>
+SIMD_TARGET inline uint8x16_t undelta(uint8x16_t a, uint8x16_t b)
+{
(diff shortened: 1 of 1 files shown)


zeux:
As a followup, this was enabled for Windows-on-ARM builds as well in https://github.com/zeux/meshoptimizer/commit/bab8b98a58c20d6a2c4511bd6f23912a6984d337

PR #830, opened 2025-01-06
Update WASI SDK to 25

zeux:
This effectively upgrades clang from 18 to 19.

As part of this, gltfpack.wasm now *requires* support for reference-types; this breaks Node 16 support.

Because gltfpack native builds are always recommended over Wasm builds, and Node 16 was end-of-lifed in September 2023, I will begrudgingly bump the requirements. My general policy is to support versions that go *further* back than this, but you should basically always use gltfpack native in production, and since this is an upstream issue resolving which will make gltfpack.wasm maintenance even harder, I am not going to do that absent feedback from actual production users who are still on Node 16.


zeux:
`-mcpu=mvp` is broken on LLVM 19; the produced gltfpack Wasm binary contains uses of reference types, and `-mno-reference-types` does not disable these:

```
 0379de: 28 02 00                   |     i32.load 2 0
 0379e1: 28 02 1c                   |     i32.load 2 28
 0379e4: 11 87 80 80 80 00 80 80 80 |     call_indirect 0 (type 7)
 0379ed: 80 00                      |       
 0379ef: 02 40                      |     block
 0379f1: 20 03                      |       local.get 3
```

zeux:
Looks like less of an LLVM problem and more of a wasi-libc problem: the overlong call_indirect encodings come from various parts of standard library, like exception support code - `__cxxabiv1::__class_type_info::can_catch` - or `fclose` when `-fno-exceptions` `-fno-rtti` is specified. This is a blocker for the upgrade unless gltfpack was to drop compatibility with earlier versions of Node.

zeux:
Filed upstream issue for now https://github.com/zeux/meshoptimizer/pull/830

jedisct1:
You can use [Zig](https://ziglang.org) instead of a precompiled version of wasi-libc such as the wasi-sdk package.

```sh
zig cc -Os --target=wasm32-wasi -mcpu=baseline
```

compiles code for the WebAssembly MVP. 

For newer versions of Node or other runtimes, features can be selectively added the same way, for example:

```sh
zig cc -Os --target=wasm32-wasi -mcpu=baseline+simd128+bulk_memory
```

The wasi libc which is recompiled on the fly with the same optimization and target flags as the rest of the application, so you always get better code than from a single precompiled package of the wasi libc.

jedisct1:
Does Meshoptimizer even need WASI at all? Looks like it doesn't require I/O access, so you maybe you could even compile to freestanding WebAssembly. That would work everywhere, including in web browsers without any dependencies.

Just replace `wasm32-wasi` with `wasm32-freestanding`.

zeux:
There's a difference between meshoptimizer modules and gltfpack:

- meshoptimizer modules do not require anything but the allocator; it looks like that happens to work with new libc as well, maybe by chance.
- gltfpack requires I/O via WASI, hence it needs the full libc to work.

The change in support would be just for the latter.

zeux:
... well, `wasm32-freestanding` doesn't work anyway as it can't discover CRT headers like assert.h, but the code for modules is compiled with `-nostartfiles -nostdlib` so it should be isolated from any libc binaries.

PR #841, opened 2025-02-13
Experimental cluster partitioning

zeux:
This change introduces a new experimental algorithm,
`meshopt_partitionClusters`, which partitions the input clusters into
groups of approximately equal sizes, grouping neighboring clusters using
implied vertex-based adjacency and prioritizing groups that reduce
boundary vertices. This is intended as a replacement to METIS cluster
partitioning for hierarchical simplification pipelines.

The input to the algorithm is an index buffer for each cluster; the adjacency
is automatically inferred from this data. For best results on meshes with
vertex splits it's recommended to supply position-welded ("shadow") indices,
which can be generated via `meshopt_generateShadowIndexBuffer` on the
array before partitioning (see `main.cpp`/`simplifyClusters` example), or by
manually remapping the indices based on a vertex remap (see `nanite.cpp`/
`partition` example).

Note that right now, the algorithm does not use spatial proximity for
grouping considerations, and only groups clusters with shared vertices;
so on some meshes, single-cluster groups may be common for separate
small connected components. In this situation, METIS will group these
clusters pretty arbitrarily which isn't obviously better than leaving them
ungrouped. In both algorithms, adding fake links based on positional
data between clusters should fix this issue but in the future meshopt
algorithm will likely start taking vertex positions as well to automatically
mitigate that.

For the structure of the algorithm, three options were evaluated: greedy
walk (similar to clusterizer) with the extra feature of appending
isolated clusters to adjacent groups; parallel grow which started with a
set of seed clusters for each connected component and grew each group
starting from seeds; and greedy merge which keeps a sorted set of
unfinished groups and iteratively picks the smallest group and tries to
merge it with the best additional group. The merge algorithm generally
produced best results; the specific heuristics here matter as well, and after
a lot of experimentation the current sorting heuristic is just based on
total vertex count (so smaller clusters get merged first), and merge priority
is trying to maximize shared vertices between merged groups and to
merge with smaller groups as well. The heuristics are subject to change in
the future.

When used in a hierarchical pipeline, the new algorithm seems to generally
be similar or slightly better vs METIS in terms of resulting DAG quality and is
a little faster to compute the partition. The test results were mostly produced
using METIS to generate clusters, as meshopt clusterizer is currently known to
generate disconnected clusters and the number of disconnected clusters at
various LOD levels was a key comparison metric. That said, using meshopt
clusterizer with this partitioner also seems to produce slightly better results
vs meshopt clusterizer with METIS partitioner.

The target group size provided to the algorithm is a rough guideline; the
resulting partitions may be smaller or larger, and the group size logic might
require some adjustments in the future. The partitioner doesn't try too hard
to keep the partition sizes uniform, which is key for having a simpler algorithm
work.

Several of the potential future work items (positions as input, maybe more
targeted group size control) would require changing the interface, so just as
with all other experimental algorithms, the interface stability or behavior is
not guaranteed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -29,6 +29,7 @@ set(SOURCES
     src/indexgenerator.cpp
     src/overdrawanalyzer.cpp
     src/overdrawoptimizer.cpp
+    src/partition.cpp
     src/quantization.cpp
     src/simplifier.cpp
     src/spatialorder.cpp
--- a/demo/main.cpp
+++ b/demo/main.cpp
@@ -657,12 +657,13 @@ void simplifyComplete(const Mesh& mesh)
 
 void simplifyClusters(const Mesh& mesh, float threshold = 0.2f)
 {
-	// note: we use clusters that are larger than normal to give simplifier room to work; in practice you'd use cluster groups merged from smaller clusters and build a cluster DAG
-	const size_t max_vertices = 255;
-	const size_t max_triangles = 512;
+	const size_t max_vertices = 64;
+	const size_t max_triangles = 64;
+	const size_t target_group_size = 8;
 
 	double start = timestamp();
 
+	// build clusters (meshlets) out of the mesh
 	size_t max_meshlets = meshopt_buildMeshletsBound(mesh.indices.size(), max_vertices, max_triangles);
 	std::vector<meshopt_Meshlet> meshlets(max_meshlets);
 	std::vector<unsigned int> meshlet_vertices(max_meshlets * max_vertices);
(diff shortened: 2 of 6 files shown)


JMS55:
I'd be interested in a comparison between this, my METIS-based code in Bevy (which besides spatial weights for grouping (TODO) I feel is pretty solid now), and nvidia's new library which uses SAH splitting (more optimal for RT probably).

I sadly don't have time to go hooking each of those up to Bevy, but maybe that's something you're interested in doing for meshopt.

zeux:
I considered testing this in Bevy directly, but the current situation with meshlet processing made me give up fairly quickly because it was difficult to hook it up; I tried to do this in early January, hit some issues in the asset system that were difficult to debug and went elsewhere. However, Bevy's use of METIS is fairly close to `demo/nanite.cpp` in this project, so I'd expect similar results.

I plan to look into RT focused clusterization & partitioning separately; both partitioning and clusterization really need to be able to optimize based on axis aligned spatial bounds over *any* other criteria for that use case as far as I understand the tradeoffs. So I'd expect NV library to work better for that use case for now, although it has surprising corner cases, e.g. it can refuse to process a cluster with <256 vertices (which is a requirement for new cluster based RT APIs) for some meshes.

JMS55:
Gotcha, totally understandable. It's very much a big imperative function with a lot of bitpacking and not much abstraction at the moment, as feature development has taken priority over cleanup of the from_mesh code.

If you're interested in trying to hook this up to Bevy again though, let me know. I would be happy to help figure out the issues you ran into and get meshopt's clusterizer hooked up (I just don't have time to write all the code myself, currently working on BVH-based culling and misc other Bevy things like RT GI).

PR #851, opened 2025-03-10
clusterizer: Implement global meshlet flow

zeux:
Up until now, the clusterizer made all decisions about how to continue and restart the meshlets based on local information - either current meshlet (for continuation) or previous meshlet (for restart). On large meshes, this
often resulted in a meandering traversal that left gaps in the mesh. These gaps would need to be filled later;
because the gaps had uneven sizes, this could result in disconnected clusters.

This change introduces global flow: the starting triangle for meshlets is now selected to prioritize a particular global traversal of the mesh. This is based on sorting by distance to a specific anchor point (arbitrarily chosen to be the negative corner of the bounding box), as well as by the sum of live counts which was the restart metric before this change. Both are important: distance sorting results in forcing meshlets to cover gaps in the mesh earlier which reduces the chance of disconnects, while live sorting results in a much cleaner meshlet fill locally.

Because of the live sorting, we can't easily use the KD tree (also, since we don't remove nodes from a KD tree, quering the same point will become progressively slower as more and more triangles would need to be skipped). This technically turns the sort above in a O(N) operation; if done before every meshlet, the entire process becomes O(N^2) and unusably slow for large meshes.

Instead, we maintain a set of triangle seeds of a limited size, and add a few seeds after finishing every meshlet, minimizing the metric above. Some corners are cut for performance, such as just selecting a single neighbor triangle per vertex and using a simpler replacement logic. The set is re-scored when starting every meshlet; this needs to be done for live triangles as we don't maintain that metric per triangle in an incremental fashion; however, since the set is of a limited size, the entire process stays linear and the performance degradation for meshlet generation is minimal (<1%).

This results in a significant improvement in cluster disconnections in various meshlet configurations (note, while 1.0 is the theoretical optimum, on a few of these meshes the mesh has many small disjoint features which makes 1.0 impossible to ever reach):

image

Reducing cluster splits also results in a small reduction in boundary size (which slightly improves vertex sharing and reduces locked edges in clustered simplification) and occasionally a small reduction in overall meshlet count. Testing the rasterization performance on geometry dense scenes with various cluster culling optimizations yields a small runtime speedup (3-5% depending on the mesh and GPU, AMD/NV were tested). Raytracing performance seems to be affected to a smaller degree, because the test harness uses more aggressive "flex" setup than the chart above, but the overall number of meshlets is also slightly reduced because fewer of them need to be split.

### DIFF
--- a/demo/main.cpp
+++ b/demo/main.cpp
@@ -1484,6 +1484,7 @@ void processDev(const char* path)
 
 	bool dump = getenv("DUMP") && atoi(getenv("DUMP"));
 
+	meshlets(mesh, /* scan= */ false, /* uniform= */ false, /* flex= */ false);
 	meshlets(mesh, /* scan= */ false, /* uniform= */ true, /* flex= */ false);
 	meshlets(mesh, /* scan= */ false, /* uniform= */ true, /* flex= */ true, dump);
 }
--- a/src/clusterizer.cpp
+++ b/src/clusterizer.cpp
@@ -366,7 +366,7 @@ static bool appendMeshlet(meshopt_Meshlet& meshlet, unsigned int a, unsigned int
 	return result;
 }
 
-static unsigned int getNeighborTriangle(const meshopt_Meshlet& meshlet, const Cone* meshlet_cone, const unsigned int* meshlet_vertices, const unsigned int* indices, const TriangleAdjacency2& adjacency, const Cone* triangles, const unsigned int* live_triangles, const unsigned char* used, float meshlet_expected_radius, float cone_weight)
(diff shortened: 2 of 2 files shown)


zeux:
For future me: I've experimented with more precise replacement criteria, including https://github.com/zeux/meshoptimizer/commit/506f03c and a fully precise 4 element top-n; unfortunately, they are all improving the results only ever-so-slightly in aggregate, and making results worse on a couple meshes where it would be nice to not regress. This might be due to some properties of the metric that I don't fully understand, or due to the metric just being suboptimal; it's possible that a better global metric will be able to take better advantage of this, but for now it's probably best to leave this as is even though the replacement is somewhat ad-hoc in certain cases.

PR #881, opened 2025-04-29
simplifier: Add experimental simplifyPrune function

zeux:
When simplifying meshes, SimplifyPrune flag is useful to be able to
discard connected components that are below the currently tracked error.

In some cases, this gives insufficient control over the result. Notably,
it might be desirable to remove components below an acceptable size
cutoff which is larger than the simplification error; additionally, it
can be useful to remove components *without* doing any other types of
simplification - this can be used for mesh cleanup, especially after
isosurface extraction.

This change exposes the existing component pruning functionality using a
separate explicit function call, in addition to the simplification flag.


zeux:
Example of using this for isosurface extraction cleanup:

Screenshot From 2025-04-29 16-01-32

PR #906, opened 2025-06-30
Link Threads::Threads instead of pthread

toge:
There is another improvement I'd like to see in CMakeLists.txt.

Even on UNIX systems, some environments don't provide pthreads, so I would prefer using Threads::Threads instead of pthreads.
This allows linking with C++11+ Threads in a platform-independent manner, supporting a broader range of environments. (ex. Android NDK)


zeux:
Thanks!

PR #907, opened 2025-07-01
 simplifier: Account for triangle filtering when finalizing grid size

### LINKED ISSUE #133: simplifySloppy overestimating actual triangle count
Hi, I'm working on a project which includes automatically simplifying arbitrary meshes to various triangle counts. I've been using simplifySloppy because it seems more reliable for various poor quality input meshes, and being able to achieve very low triangle counts for low LODs.

I found that when reducing a highly detailed mesh to very low triangle counts, the grid size being used sometimes ended up being bigger than it needed to be - this is because countTriangles() wasn't taking into account that some triangles become duplicates and then get eliminated during filterTriangles(), after quantization.

I've made a fix for it here: https://github.com/virtalis/meshoptimizer/commit/62ab34f2b4bf23e9d8f26359284cddfc20500b7d. Would you like me to open a pull request, or was it left this way intentionally for speed? It now requires creating a hash table of triangles during each loop iteration at the start of the algorithm, which will obviously slow it down a bit compared to the simple sum it was doing before.

--

I've also made some changes to prevent vertices being collapsed when their normals are drastically different, by expanding the grid into an additional 3 dimensions for normal space: https://github.com/virtalis/meshoptimizer/commits/simplifySloppy-normals. I'm not sure if these are changes you'd be interested in, but I can open another issue to discuss them if you are.

zeux:
Binary search for the target grid size is performed using an approximate
calculation for triangle count which overestimates the result as it does
not filter out duplicate triangles. This often creates cases where a
slightly larger grid would still fit in the triangle budget.

This could be solved by using post-filtered triangle count during the
binary search but that's very expensive. Most often, the excess is small
enough that we can only bump the grid size by 1, so instead we perform a
single probe pass to check if increasing the grid size is possible.

To do this we use the same code that we use for triangle filtering, but
use vertex ids instead of vertex cells; it is still correct because
there's a 1-1 relationship between these, and while the destination
buffer becomes nonsensical briefly, it's only used as temporary storage
during hashing.

Fixes #133.


zeux:
Need to rethink this after further testing…

PR #914, opened 2025-07-14
simplifier: Use point quadrics for regularization

### LINKED ISSUE #664: Simplification should prefer more uniformly tessellated meshes
Hello, below is a video of a 100 vertices plane simplified with ratios [0.5, 0.25, 0.125, ...] and an error of 1

It seems like the simplification is suddenly stopped once a threshold is reached where areas of the mesh then stay untouched.
We also observed this on a relatively high density sphere where it resulted of dense islands

Expected behavior
Simplification is uniformly applied

[assets.zip](https://github.com/zeux/meshoptimizer/files/14619690/simplify.zip)

https://github.com/zeux/meshoptimizer/assets/5083203/24eb2275-da32-4312-bfce-1471a197e7d0

zeux:
In addition to face and edge quadrics, we now introduce point quadrics,
that encode distance to the vertex position. We used to use these for
point simplification but that was replaced with quadric-less metric for
simplicity instead, but they can be used to improve mesh simplification.

For planar regions, regular triangle quadrics evaluate to 0 for any
point on the plane. This leads to chaotic triangulation as the collapse
order is effectively random, and can produce skinny triangles which is
suboptimal for rasterization performance. Additionally, quadrics like
this can't be minimized.

Adding point quadrics to vertex quadrics with a very small factor
improves both issues significantly. The factor is chosen to minimally
affect the overall simplification error, but still improve triangulation in
planar areas.

Additionally, a new experimental option, `meshopt_SimplifyRegularize`,
uses a higher regularization factor. This results in more uniform triangulation
even in non-planar areas, and trades off some geometric/attribute quality
instead. This is useful for cases when objects are deformed at runtime, like
skinning, and may improve the overall shape in some other cases while
reducing the quality of smaller details. The error is incorporated into
target/result error which means the LOD selection behavior will adjust
accordingly (which is also helpful for deformable objects).

Fixes #664.


zeux:
Before this PR:

Screenshot From 2025-07-14 10-27-19

This PR, default settings:

Screenshot From 2025-07-14 10-27-29

This PR, `meshopt_SimplifyRegularize`:

Screenshot From 2025-07-14 10-33-21

(triangle counts are the same in all of these)

PR #919, opened 2025-07-24
simplifier: Classify half-seams as seams instead of borders

zeux:
In complex meshes it's possible to encounter a somewhat unusual type of vertex: one that only has a single copy (wedge), and that has one incoming and one outgoing open edge, but both edges connect to vertices with the same position.

Previously, we classified vertices like this as border, because border classification ignored this possibility. This created cases where a border vertex is adjacent to a seam vertex, but neither can be collapsed onto each other because of collapse restrictions.

Doing the right thing here is somewhat complicated. Locking vertices like this to avoid misleading classification regresses cases where a half-seam vertex like this is connected to a locked vertex; these are the cases that in theory may lead to issues on discontinuities but in practice we don't seem to observe them. Classifying this as a new type does not solve that either, as deciding that this vertex can move leads to similar issues in theory.

For now we re-classify vertices like this as seams. The seam handling logic is already safe in case a seam vertex only has a single copy, modulo an assertion that this change adjusts. If this leads to problems down the line, the next best thing is to lock these, at least until we have new complex collapse rules in place.

### DIFF
--- a/src/simplifier.cpp
+++ b/src/simplifier.cpp
@@ -394,6 +394,13 @@ static void classifyVertices(unsigned char* result, unsigned int* loop, unsigned
 				{
 					result[i] = Kind_Manifold;
 				}
+				else if (openi != ~0u && openo != ~0u && remap[openi] == remap[openo] && openi != i)
+				{
+					// classify half-seams as seams (the branch below would mis-classify them as borders)
+					// half-seam is a single vertex that connects to both vertices of a potential seam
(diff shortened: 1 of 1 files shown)


zeux:
An example of something like this for an extreme simplification; here the seam is a normal seam, and orange color is used to annotate problematic edges that used to be border-seam and were effectively locked, and are now seam-seam and are eligible for a collapse.

before | after
----------|-------
image | image

PR #921, opened 2025-07-30
simplifier: Implement experimental position/attribute update

zeux:
This change introduces a new function, `meshopt_simplifyWithUpdate`, which computes optimal (from the appearance point of view) positions and attributes for the resulting mesh. The resulting values are written to the input `vertex_positions` / `vertex_attributes` buffer, destructively modifying it for the resulting vertices; accordingly, the function also modifies the `indices` buffer in place.

After this update, the vertex buffer naturally can't be used for the original (denser) LOD. It's expected that this is used with with subsequent compaction (via `meshopt_optimizeVertexFetch`), or in clustered simplification use cases, where multiple calls to `meshopt_simplifyWithUpdate` with `meshopt_SimplifySparse` flag can be made and the updated cluster vertex data can be extracted. If the original data was important it should be copied before the call.

Since the quality gains are more apparent on more significant reduction, it might be an interesting idea to use non-updated vertices for the first one or two levels in the LOD chain, and then switch to updated vertices; this limits the extra memory use by duplicate vertex data.

For best results, unlike `simplifyWithAttributes` where UV inputs are typically optional, the simplifier needs to have access to, and have non-zero weights on, texture coordinate data to be able to reproject it after positions move (alternatively it might be useful to only use normal data and keep original positions - the update function retains the original indexing so this is easy to do after calling the function). Attributes like normals and colors may need to be renormalized or clamped after the function returns new data. Attributes like bone indices/weights don't need to be communicated for reasonable results (but regularization via `meshopt_SimplifyRegularize` may still be helpful).

Additionally, when computing LOD chains, it's recommended to compute the next LOD from the previous LOD in the chain. This improves attribute appearance even without the update and has been highlighted in the documentation.

This is highly experimental: the interface and behavior of the function may change, and there might be occasional remaining issues where the appearance regresses (or does not improve enough) after the update. Currently, the update does not affect the returned error, so the LOD selection and simplification stays consistent between the options - this, among other things, may change in the future.

Example 2M => 10k simplification (0.5%) without an update (one shot) vs with (iterative in 1/2 steps):

image


castano:
It may be worth checking the position of the new vertex when using the volume conservation constraint. If the new vertex is far from the original vertices it may be worth falling back to the unconstrained position.

Volume conservation was often an improvement for us, but once we reached a certain simplification level, the results were less than ideal. Imagine a long parallelepiped, if you removed any of the vertices the volume would go down, but to make up for that the volume conservation would displace the vertex well outside the original shape.

When I implemented this I was hoping to find a way to treat the volume gradient as another attribute to minimize, instead of a hard constraint, but wasn't able to figure out how.

castano:
BTW, this is looking great. So much better than my implementation.

zeux:
I ended up implementing some additional boundary checks and tweaks in https://github.com/zeux/meshoptimizer/pull/926 - they won't fully prevent the issue you're talking about, but they help (although they abandon the solve instead of falling back to unconstrained - the problem with unconstrained is it potentially brings back volume loss... I also tried solving the vertex quadric in isolation but that yielded roughly the same results as solving attribute quadric with volume constraint so I don't use that yet). I also looked a little bit at using the Lagrange multiplier to understand if the volume constraint is pushing the vertex too far but this didn't work super well. It's also possible to use the vertex-only quadric to sanitize the results - this wasn't necessary in my testing so far but it remains the option (just as a separate solve) since I have it stored separately.

castano:
Ah, yeah, that should prevent those vertices from collapsing, which would avoid the issue. In my case it was beneficial to allow some volume loss. I had a pass that eliminated connected components based on volume & surface and I run it before and after simplification. So, some components that were not eliminated during the first pass would go away on the second.

The way my simplifier worked is quiet different, though. I measured the quadrics based on the updated vertex position, while I think you update the vertex position after you have chosen the vertex collapse, that is, the quadric minimization does not affect the ranking. I think what I did is the standard approach, it's certainly more expensive, but not sure how much does it matter in practice.

Anyway, this is based on a superficial reading of the code, so I may have missed something. I wanna find some time to read it in more detail.

zeux:
Yes - it's not quite the standard approach. I don't plan to rank based on solved data, but in the future solve will be a bit more incremental and I plan to try using optimal attribute error (aka attribute error based on the optimal attribute values) which will probably approximate a different way to rank better.

PR #930, opened 2025-08-13
simplifier: Implement experimental permissive mode

zeux:
When a new setting, `meshopt_SimplifyPermissive`, is enabled, the
simplifier is less restrictive around attribute discontinuities:
vertices that were on seams, incollapsible seams, or vertices with more
than two wedges, are marked as Complex. Collapsing those may break
attribute discontinuities, guided by the error; in this mode, it's still
possible to protect some discontinuities like UV seams by using external
classification: bit mask 2 (bit 1) in `vertex_lock` can be used to protect
the discontinuity, effectively retaining existing classification.

The internal behavior for this  is going to change somewhat in the future;
currently, border vertices with more than one wedge will be tagged as
complex which can distort geometric borders. There might be other cases
where we shouldn't tag unprotected vertices as Complex.

As an example, you might want to protect vertices that lie on a UV discontinuity
by checking if the UV of each vertex matches the UV of the first vertex with
the same position, and protect vertices that lie on a sharp normal crease by
checking if the angle between the normal of each vertex and the first vertex
with the same position is above a threshold. In the future, something like this
could be provided as a separate function to call beforehand, but it's not too
difficult to implement (see `demo/simplify.html` as an example).

> [!WARNING]
> This functionality is very experimental. While I suspect the basic API and behavior
> will stick around, the details of what vertices will stay locked if unprotected,
> how the unprotected vertices can move and others will likely change in the future.
> This is provided as an early preview for the upcoming release to gather feedback.

### DIFF
--- a/demo/nanite.cpp
+++ b/demo/nanite.cpp
@@ -69,6 +69,7 @@ const bool kUseLocks = true;
 const bool kUseNormals = true;
 const bool kUseRetry = true;
 const bool kUseSpatial = false;
+const bool kUsePermissiveFallback = true;
 const bool kUseSloppyFallback = false;
 const int kMetisSlop = 2;
 const float kSimplifyThreshold = 0.85f;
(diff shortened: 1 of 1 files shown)


zeux:
This can allow to simplify meshes that would previously get stuck very early:

before: stuck at ~90% | after: at 25% | after: at 2.6%
----|------|-----
Screenshot 2025-08-11 at 9 06 54 PM | Screenshot 2025-08-11 at 9 08 22 PM |  Screenshot 2025-08-11 at 9 07 47 PM

As well as improve simplification quality on meshes where parts were locked
due to faceted shading, which matters a lot for significant reduction:

before, stuck at ~3% | after, at 3% | after, error limited at 1%
-------|-------|-----
Screenshot 2025-08-12 at 5 10 01 PM | Screenshot 2025-08-12 at 5 10 32 PM | Screenshot 2025-08-12 at 5 10 52 PM

PR #941, opened 2025-08-25
clusterizer: Remove negative cone_weight from buildMeshletsFlex

zeux:
The original idea has been to change the heuristic to produce higher
quality splits for raytracing focused workflows. Unfortunately, this
doesn't *really* work: the impact on RT efficiency is, on average, net
negative, and in general this doesn't solve the problem very well even
for meshes where it is an improvement. Since a new algorithm,
buildMeshletsSpatial, solves that use case much better, we don't need
this feature anymore.

Since buildMeshletsFlex is experimental, this can be done without
additional versioning. If we need some sort of adjustment like this in
the future, we can still use negative cone_weight as a flag, it's just
that this specific heuristic does not work very well.

While buildMeshletsFlex itself is still useful as it provides better control
over spatial locality, and it is used by some applications, negative cone
weight is not used by any application I'm aware of so the removal should
hopefully go unnoticed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -211,7 +211,7 @@ meshlets.resize(meshlet_count);
 
 Depending on the application, other strategies of storing the data can be useful; for example, `meshlet_vertices` serves as indices into the original vertex buffer but it might be worthwhile to generate a mini vertex buffer for each meshlet to remove the extra indirection when accessing vertex data, or it might be desirable to compress vertex data as vertices in each meshlet are likely to be very spatially coherent.
 
-For optimal performance, it is recommended to further optimize each meshlet in isolation for better triangle and vertex locality by calling `meshopt_optimizeMeshlet` on vertex and index data like so:
+For optimal rasterization performance, it is recommended to further optimize each meshlet in isolation for better triangle and vertex locality by calling `meshopt_optimizeMeshlet` on vertex and index data like so:
 
 ```c++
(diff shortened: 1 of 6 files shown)


zeux:
Some data to support this decision:

- Based on SAH costs, the AA heuristic produces anywhere from -5% to 5% SAH costs on a variety of test meshes, but there are more meshes where it regresses vs improves:

plot-aa

- The geomean of SAH improvement is negative in all triangle count buckets:

 bucket_pow10 |              bucket_range |  count | geomean_impr_pct
-|-|-|-
2 |             [100, 1,000)      |7  |         -1.827%
3   |        [1,000, 10,000)     |38 |          -0.921%
4     |    [10,000, 100,000)    | 33  |         -0.248%
5     | [100,000, 1,000,000)    | 19   |        -0.827%
6  | [1,000,000, 10,000,000)     | 8     |      -0.025%
7 | [10,000,000, 100,000,000) |     1     |      -5.810%

- On RTX 5070, performance is similarly highly variable; while it improves performance on some meshes a little bit, it also degrades it on other meshes. Generally speaking, RT performance fluctuations here are small - compared to using the spatial algorithm which significantly improves RT performance - and on some meshes, there's noticeable degradation when compared against cone_weight = 0.

This can be compared to the other feature of flex clusterization, split_factor, which sees clear improvement in SAH costs (and geomean improvement of 2-5% depending on the bucket):

plot-sf

PR #944, opened 2025-08-26
Validate output extension up front to prevent late failure and unnecessary work

### LINKED ISSUE #943: Fail fast on invalid output extension to avoid wasted work and stray texture copies
The output filename extension in gltfpack is currently validated only at the very end of processing.
If an unsupported extension (anything other than .gltf or .glb) is provided, the tool still:

- parses and transforms the full scene,
- encodes and copies textures into the target directory,
- and only then aborts with Error saving …: unknown extension (expected .gltf or .glb).

This behavior leads to wasted processing time and side effects (e.g. copied textures) for an operation that is guaranteed to fail.

**Expected Behavior**

Validation of the output extension should occur immediately after parsing the -o argument.
If the extension is not .gltf or .glb, the tool should exit with the same non-zero code (4) that is used by the late check today.

**Benefits**
- Faster feedback to users when providing an invalid extension.
- No unnecessary processing or texture copies.
- Clearer and more predictable CLI behavior.

ynnob:
This PR to solve #943  adds an early validation of the output filename extension in gltfpack(). 
If -o is provided and the extension isn’t .gltf or .glb, the tool now exits immediately with the same non-zero code used by the current late check. 

This avoids:
- running the full processing pipeline on an output that will be rejected,
- encoding/copying textures into the target directory,
- confusing partial side-effects followed by a late error.

Tests (-test) that call gltfpack(path, NULL, …) continue to work as before. 

Fixes #943.

### DIFF
--- a/gltf/gltfpack.cpp
+++ b/gltf/gltfpack.cpp
@@ -1010,6 +1010,15 @@ int gltfpack(const char* input, const char* output, const char* report, Settings
 	std::string iext = getExtension(input);
 	std::string oext = output ? getExtension(output) : "";
 
+	if (output)
+	{
+		if (oext != ".gltf" && oext != ".glb")
+		{
+			fprintf(stderr, "Error: unsupported output extension '%s' (expected .gltf or .glb)\n", oext.c_str());
+			return 4;
+		}
+	}
+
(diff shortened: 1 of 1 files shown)


zeux:
Thanks!

PR #954, opened 2025-09-17
demo: Rework clustered LOD example code

zeux:
This change significantly reworks the clustered LOD builder example. The actual implementation is mostly identical in the flow, however it's structured to be easier to start with or reuse when implementing a system like Nanite:

- Removed extraneous prototype code, including METIS dependency
- Expose a proper C interface with a generalized input mesh
- Use permissive simplification with sloppy simplification as fallback (but expose a lot of configuration knobs)
- Support raster-focused and RT-focused builds

To stress test this, this PR also includes `tools/clodgltf.cpp` which can be built separately and used to process the entire NVidia Zorah sample scene (~1.7B triangles) to generate a full DAG for every mesh. This process takes a little under 3 minutes on Ryzen 7950X using 16 threads and ~60 GB of memory. There's no processing or serialization of output data at the moment; this is mostly used to stress test the rest of the code and may be removed in the future.

The actual code is still using a fair amount of STL and is not too careful with allocations. This might be good to fix in the future but this PR is already too large and GitHub lost track of the fact that the implementation got renamed, so this is probably better to push as is as a checkpoint :)


zeux:
Having some second thoughts around some parts of this. Notably, the error accounting for clusters was previously more intuitive and robust in absence of explicit groups; stuck clusters would automatically render at correct requested depth etc.

A better first step would be to pull apart a few commits here to improve the existing demo; some things like permissive-by-default and sloppy integration are great as is, and so is the general cleanup. Will close this for now and open a different PR tomorrow.

PR #989, opened 2025-11-29
 simplifier: Use parallel collapses for complex vertices based on loops

zeux:
Collapsing all wedges into a single wedge for complex-complex collapses
is simple but results in high-error collapses alongside attribute
discontinuities. This is a problem because it either results in poor
attribute quality, or in poor geometric quality elsewhere as the
simplifier is forced to look for lower error collapses.

This change instead adjusts complex collapses to use parallel collapses
if possible: wedges are collapsed individually towards different wedges
along the original edges. This makes complex collapses work similarly to
seam collapses if possible.

To implement this without extra performance issues, we use loop
metadata. Because loops are really tracking a single incoming/outgoing
half-edge in the original (unwelded) topology, a typical attribute
discontinuity will have X wedges (seams only have 2), and each of the
wedges will have their own loop/loopback that correctly trace the
adjoining triangle boundary. Thus we can use these to guide the
collapses.

In some complex topological situations this may not always find the
correct wedge target, but fundamentally complex-complex collapses may
*not* have a correct wedge target in all cases (if a discontinuity has 3
wedges then any collapse will have to collapse one of them onto a wedge
that was not connected to the original). In the future we could use
error or position patching to resolve these cases but that requires
invasive changes with potentially problematic consequences, whereas
using loop metadata should be cheap and safe.

This has no effect on performance on most meshes (and is only applied
for permissive mode); on meshes where almost every vertex is
complex, this costs under 10% performance but results in much better
quality. In general, after this "soft" protection like normal creases is
mostly unnecessary because the simplifier will do the right thing while
it's possible. Protecting UV edges may still be important to avoid UV
distortion, although even without UV protection the quality can remain
quite high if the UVs are weighted properly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -503,7 +503,7 @@ lod.resize(meshopt_simplifyWithAttributes(&lod[0], indices, index_count, &vertic
     target_index_count, target_error, /* options= */ meshopt_SimplifyPermissive, &lod_error));
 ```
 
-To maintain appearance, it's highly recommended to use this option together with attribute-aware simplification, as shown above, as it allows the simplifier to maintain attribute appearance. In this mode, it is often desirable to selectively preserve certain attribute seams, such as UV seams or sharp creases. This can be achieved by using the `vertex_lock` array with flag `meshopt_SimplifyVertex_Protect` set for individual vertices to protect specific discontinuities. To fill this array, use `meshopt_generatePositionRemap` to create a mapping table for vertices with identical positions, and then compare each vertex to the remapped vertex to determine which attributes are different:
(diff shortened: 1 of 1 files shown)


zeux:
Example of how this helps on a completely faceted mesh (smaller error & better shading):

before | after
----------|-------
image | image

Example of how this helps alongside UV seams with high texture weight (high error collapses weren't taken before):

before | after
----------|----------
Pasted image 20251128131331 | Pasted image 20251128131440

Example of how this helps alongside UV seams with *zero* texture weight:

before | after
----------|-------
image | image

With this, permissive mode with enabled protection on UV seams should generally not be worse wrt quality or appearance than not using it, and in some cases it may even be practical to disable UV seam protection, as long as simplifier *does* have access to UV data with carefully setup weight - the good results in the last comparison are not guaranteed otherwise :) Accordingly it should be unnecessary to protect normal creases so the documentation was updated accordingly.

Permissive mode will remain experimental because there's still cases where behavior is not ideal and there's a few different ways to improve this that may be risky.

PR #997, opened 2025-12-09
js: Fix resolution of TypeScript .d.ts definitions

donmccurdy:
For TypeScript projects that have adopted certain newer configurations (`"moduleResolution": "nodenext"` in tsconfig.json is probably the most common), it's necessary to include .js extensions on import paths within `.d.ts` files. (Despite the fact that we're referring to other `.d.ts` files, yeah...). Fortunately, this is backwards-compatible with older TypeScript configurations.

~~The addition of the `package.json#typesVersions` fixes a build issue for glTF Transform after I switched over to importing from subpaths like `meshoptimizer/decoder`, but I'm still trying to figure out why — I don't typically use the "typesVersions" field so this was unexpected to me. Duplicating the .d.ts paths under both `exports` and `typesVersions` shouldn't be necessary.~~ Reverted this part. 

I'll mark this as a draft until I understand that better. I suspect this is not a common issue for JS/TS projects, so no particular rush on reviewing or publishing a patch. :)

Background:

- https://github.com/donmccurdy/glTF-Transform/pull/827

### DIFF
--- a/js/package.json
+++ b/js/package.json
@@ -1,6 +1,6 @@
 {
 	"name": "meshoptimizer",
-	"version": "1.0.0",
+	"version": "1.0.1",
 	"description": "Mesh optimization library that makes meshes smaller and faster to render",
 	"author": "Arseny Kapoulkine",
 	"license": "MIT",


donmccurdy:
I think what's left in the PR should be ready, for the moduleResolution=NodeNext fix!

donmccurdy:
I don't expect any further issues with the 'meshoptimizer' root path import after this fix, correct!

For the `meshoptimizer/*` subpath imports, I think it's probably not Rollup broadly, but the Rollup plugin used to resolve types (if any). In my case that's rollup-plugin-typescript2, used in Microbundle. Based on a couple issues...

- https://github.com/ezolenko/rollup-plugin-typescript2/issues/437
- https://github.com/developit/microbundle/issues/1094#issuecomment-3407740352

... it looks like my build tool doesn't fully support `package.json#exports`, so I should just be using the root path import instead (or updating the build configuration).

I might try to (1) publish a glTF Transform v4.3 release in the next few weeks, and then follow up with a v4.4 release updating the build to support package.json#exports and updating to meshoptimizer v1.0.0. At that point it should be compatible with the current meshoptimizer v1.0.0 package. If you wanted to wait on publishing v1.0.1 until then, as no one else has reported the moduleResolution=NodeNext compatibility issue yet, that could be a good option too.

PR #1007, opened 2026-01-25
gltfpack: Implement initial support for XUASTC

zeux:
When compiling against latest Basis Universal (2.0+), we can now use XUASTC compression. It supports variable block sizes and a much larger gamut of quality settings from the perspective of the resulting file size, using DCT to compress the block data with arith/Zstd on top.

This format is currently not supported by any glTF extensions; as such, this might need refinements if this happens. For now, the block size is fixed at 4x4; this could be controlled via quality settings or a separate future command line argument.

The quality curve for now is somewhat approximate and based on the target size; the quality impact has not been analyzed yet.

texture_quality_chart

### DIFF
--- a/gltf/encodebasis.cpp
+++ b/gltf/encodebasis.cpp
@@ -30,24 +30,39 @@ struct BasisSettings
 	int etc1s_q;
 	int uastc_l;
 	float uastc_q;
+	int xuastc_e;
+	int xuastc_q;
 };
 
 static const BasisSettings kBasisSettings[10] = {
-    {1, 1, 0, 4.f},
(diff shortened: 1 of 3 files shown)


arpu:
looks like this is stable now?

zeux:
My understanding is that there’s no relationship defined still between XUASTC and glTF? There is no new extension and existing glTF extension only allows ETC1S/UASTC.

arpu:
You are right, looks like the specs for KHR_texture_basisu needs expanded
hopefully @lexaknyazev knows more?

PR #1043, opened 2026-04-15
 vertexcodec: Fix Wasm SIMD rotate32 to use logical instead of arithmetic shift

zeux:
i32x4_shr performs an arithmetic shift which extends the sign bit into
the portion that is then or'd with the rest of the rotate; if the
encoded high bit is 1, the output may be decoded incorrectly.

Note that this only affects Wasm and only affects cases where encoded
data uses version 1 and level 3, as xor isn't used in other cases. This
also appears to be difficult to hit on real content; the added test is synthetic.


zeux:
This was introduced in [v0.23](https://github.com/zeux/meshoptimizer/releases/tag/v0.23), a little over a year ago, and appears to be difficult to hit naturally. The issue was discovered using LLM driven static analysis - not on real data. Unfortunately, it was right after the 1.1 release, so the fix obviously didn't make it. The scope here is limited but the timeline until 1.2 is quite long (many months).

I'm considering a patch (1.1.1) release that, to reduce the release burden and minimize churn, would be scoped to meshoptimizer NPM package. This would be similar to 1.0.1 release that was also NPM only, although that one had no C++ changes.

zeux:
https://www.npmjs.com/package/meshoptimizer/v/1.1.1

PR #1046, opened 2026-04-23
tangentspace: Implement experimental tangent space generator

zeux:
This change implements `meshopt_generateTangents`, which generates per-corner tangent frames (tangent vector and orientation) following MikkTSpace algorithm.

## Motivation

`mikktspace.c` is an industry standard, but is often a bottleneck during import as it's not written for efficiency. It also has an interface that is too easy to get wrong (more than half of the engine integrations I have reviewed do not use the library correctly), a few implementation defects and some inconsistent promises about the generated data. It's also 1900 lines of code. Because engine processing pipelines typically use `meshoptimizer` for one reason or another, it would be great if they can switch to using meshoptimizer tangent generation code, and get compatible results at lower cost.

## Usage

The generated frames are *per corner*: three frames are generated for each triangle. The function supports both indexed and un-indexed inputs (similar to `generateVertexRemap` et al); the generated results do *not* depend on the indexing, but the indexed inputs are processed faster.

If you start from an unindexed input, the results of this function can be treated as another unindexed stream, and a reindexing pass will produce final vertex/index buffer.
If you start from an indexed input, it's recommended to duplicate vertices on the fly based on tangent information (see `tangents` example in `demo/main.cpp`). While you *can* omit the duplication step and simply copy tangents into existing vertex data, note that MikkTSpace algorithm splits tangents on UV mirror seams (edges where UV mapping changes winding) - such seams may not have duplicated vertices in the original indexed topology.

Compared to the reference implementation, this implementation is ~5-7x faster on unindexed inputs, ~10-11x faster on indexed inputs, and in most cases produces visually identical outputs.

## Internals

The implementation follows the MikkTSpace *paper* faithfully, and only follows the *reference implementation* (`mikktspace.c`) loosely. It is not based on the reference implementation. It is not guaranteed to produce the same results; however, short of floating point numerics and a couple rare edge cases, it should produce the same results on manifold inputs without degenerate triangles in UV space (triangles that map to a line/point).

MikkTSpace reference implementation has inconsistencies in handling of non-manifold data (it arbitrarily groups triangles into pairs, which violates order independence) and in handling of degenerate triangles (here any algorithm will have to make arbitrary choices; the choices this implementation makes may be different). In either case it should be rare that the difference materially affects the rendered result, and in these cases you'd likely see MikkTSpace generated data flip between different mesh representations anyway.

Compared to the paper, this implementation incorporates the corner weighting by projected angle; this choice of weighting is unexplained and arbitrary, but is necessary to have parity with reference implementation on well-formed meshes. This is the only significant gap in the paper on normal meshes.

Compared to the reference implementation, this implementation uses a different strategy of unifying UV-degenerate triangles with neighbors, and a different strategy of identifying adjacent triangles for non-manifold inputs. The latter uses an actual order-independent approach; the former is inherently input dependent.

The reference implementation has a few defects that this implementation fixes; neither is a problem here:

- Output tangents can be degenerate (0/0/0) in rare cases
- The angular threshold is not correctly disabled by default, which results in rare, numerics-dependent, excessive tangent splits. Notably, Blender ships a rewritten implementation that mostly follows the reference implementation but fixes this issue too.

## Future work

The current implementation has a few known defects, also present in the reference implementation:

- Output triangles can have inconsistent orientation between the three corners (+/-1)
- Output tangent vectors are not always orthogonal to the vertex normals (when using 1,0,0 fallback)
- Tangent vector weighting ignores triangle area which can lead to visibly incorrect tangents.
- Input triangles with degenerate positions but non-degenerate UVs might result in extra tangent splits

These can cause visual artifacts depending on the data representation, but they are difficult to fix in a rigorous manner. In first two cases the problems are specific to UV-degenerate triangles; the third case is dictated by reference implementation if the goal is to match it as closely as possible.

While MikkTSpace can generate both tangent & bitangent vectors separately, almost every usage in the industry uses the recommended cross product based reconstruction; hence this PR only supports that for now. It also expects the caller to figure out how to encode the tangent vectors. Additional functions could be provided to cover either use case in the future as part of the same `.cpp` file.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -42,6 +42,7 @@ set(SOURCES
     src/simplifier.cpp
     src/spatialorder.cpp
     src/stripifier.cpp
+    src/tangentspace.cpp
     src/vcacheoptimizer.cpp
     src/vertexcodec.cpp
     src/vertexfilter.cpp
--- a/demo/main.cpp
+++ b/demo/main.cpp
@@ -1388,22 +1388,74 @@ void coverage(const Mesh& mesh)
 	    cs.coverage[0] * 100, cs.coverage[1] * 100, cs.coverage[2] * 100, (end - start) * 1000);
 }
 
+struct VertexTangent
+{
+	float px, py, pz;
+	float nx, ny, nz;
+	float tu, tv;
+	float tx, ty, tz, tw;
+};
+
 void tangents(const Mesh& mesh)
 {
 	double start = timestamp();
 
 	// note: tangent generation produces a tangent vector and orientation (+-1) per *corner* (3 per triangle), even for indexed inputs
 	// you *could* then copy the tangents to the existing vertices, but that will not work correctly in case the input has UV mirroring
-	// the correct way to handle this is to deindex & reindex the mesh if you need correct results, or copy tangents to an existing vertex stream
-	// while duplicating vertices with divergent tangents on the fly.
-	// you also can pass unindexed data (indices==NULL) to meshopt_generateTangents directly, here we don't for simplicity
+	// a simple way to handle this is to work on unindexed data (indices==NULL), generate tangents, and re-index
+	// we show a more involved way that splits vertices with divergent tangents on the fly.
 	std::vector<float> tangents(mesh.indices.size() * 4);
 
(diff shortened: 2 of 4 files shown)


aras-p:
Exactly 420 lines in the implementation! Should have shipped this on April 20th :)

zeux:
Going to merge this as is for now but there will be a second PR next week with some significant improvements beyond MikkTSpace.

PR #1047, opened 2026-04-27
tangentspace: Improved tangent space generation

zeux:
This change builds on top of #1046 and improves the tangent generation behavior further. A couple fixes improve consistency with MikkTSpace while improving some situations that MikkTSpace mishandles; for example, with this change tangent frame orientation is now consistent across all three triangle corners. Additionally, by default this change switches the tangent weighting to be area-sensitive. This is a departure from MikkTSpace; on real world geometry it's often the case that MikkTSpace generates broken tangents because it over-weighs tangent on thin bevels which displaces tangents away from regular triangles.

While this is almost always an improvement, and is generally benign on other parts of the mesh that don't exhibit this problem, this technically produces different tangents which may distort the lighting if the normal maps were baked from higher resolution geometry using MikkTSpace weighting. Because of this, an option `meshopt_TangentCompatible` can be used to use the original MikkTSpace weighting.

I've also spent some time looking into how to improve tangents on degenerate UV triangles (that are zero-area in UV space), but so far all solutions have tradeoffs and aren't clearly beneficial. An experimental option `meshopt_TangentZeroFallback` can be used to zero these out which can work better as it flattens the normal map but it requires renderers to handle non-orthogonal frames properly which is atypical.

A few examples of improved weighting; the left side matches MikkTSpace perfectly whereas the right side is the new default:

compatible | default
-----------------|-------------
image | image
image | image
image | image
image | image
image | image

### DIFF
--- a/README.md
+++ b/README.md
@@ -796,6 +796,22 @@ Additionally, note that while the code above works with 32-bit OMM indices, afte
 
 When using 4-state OMMs, rasterization code produces both unknown-transparent and unknown-opaque states based on microtriangle coverage; this enables the use of forced 2-state flag during traversal for specific effects where micromap data is sufficient for reasonable quality; this is recommended for performance as this results in no any-hit invocations.
 
+### Tangent spaces
+
+Meshes that use tangent space normal maps often require per-vertex tangent vectors in addition to normals. These could be exported alongside mesh data, but this library also provides an algorithm similar to MikkTSpace that can generate them from positions, normals and texture coordinates:
+
+```c++
+std::vector<float> tangents(indices.size() * 4);
+meshopt_generateTangents(&tangents[0], &indices[0], indices.size(),
+    &vertices[0].px, vertices.size(), sizeof(Vertex), &vertices[0].nx, sizeof(Vertex), &vertices[0].tx, sizeof(Vertex));
+```
+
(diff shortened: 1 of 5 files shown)


JesseRMeyer:
For reference - Eric Lengyel mentions in FGED2 section 7.5 a maybe compatible quality addition -- to duplicate vertices on tangent frame discontinuities, set their tangents to their average if they share the same handedness and whose original tangents point in similar directions.

zeux:
This is described in the sample chapter https://foundationsofgameenginedev.com/FGED2-sample.pdf but I don't think it's very well motivated or really very compatible; it requires some sort of cutoff for directions which would need to be tuned. I have tested Lengyel's approach to implicit weighting (by averaging un-normalized tangents) last week and found it generally produces worse quality tangents, and the code doesn't handle tangent orientation correctly at all when UVs are mirrored so I think in general I am not keen on taking any FGED material about tangents into account.

Note that when the vertices are split on different attributes, e.g. if you have a second UV set for lightmaps that introduces its own seams, MikkTSpace, as well as my code, handles this correctly because the normals & UVs are identical at these vertices; FGED code doesn't and the note you mention would fix that (which is indeed correct and matches the spirit of building tangent frames that are independent of the mesh indexing).

PR #1053, opened 2026-05-20
Implement index filtering for redundant triangles

zeux:
If two vertices of a given triangle are identical after vertex shader
transform, the resulting triangle is degenerate and will not be rasterized.

By itself, this filtering is fairly straightforward to do and
applications can easily do it by themselves; however, meshes may also
contain duplicate triangles that are usually similarly redundant.

Assuming depth testing is enabled, both triangles will be rasterized and
the triangle that will be retained depends on the depth comparison
function and the triangle order (which may change after vertex cache
optimizations or meshlet conversion). In general, duplicate triangles
are likely unintentional and should not be present in the input mesh.

The new `filterIndexBuffer` function is designed to filter both kinds of
redundant triangles out, and returns a new index buffer that doesn't
contain degenerate or duplicate triangles. Positions are resolved using
binary equivalence of a vertex prefix, similarly to `generateShadowIndexBuffer`;
when skinning is used, it's important to specify both position and bone
indices/weights, as otherwise triangles that looked degenerate before
skinning could be removed even though they could become non-degenerate after.

`filterIndexBufferMulti` is a stream-based variant; instead of using a
single stream (with `vertex_size` bytes) to establish vertex equality, it
uses a set of streams to support use cases like skinning where position
should be hashed together with bone indices/weights to avoid issues
where a degenerate triangle in bind pose becomes non-degenerate after
skinning.

These technically ignore complexities around +/-0, but it's correct to
not remove some redundant triangles, just slightly less efficient.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,8 @@ When optimizing a mesh, to maximize rendering efficiency you should typically fe
 3. (optional) Overdraw optimization
 4. Vertex fetch optimization
 5. Vertex quantization
-6. (optional) Shadow indexing
+6. Index filtering
+7. (optional) Shadow indexing
 
 ### Indexing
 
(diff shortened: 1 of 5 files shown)


zeux:
The gains here are obviously situational and for clean assets filtering would be a no-op; curiously, Bistro scene has a tree that has a large number of leaf triangles duplicated. Filtering these out results in noticeable gains for both rasterization and ray tracing as long as the tree is prominent in the view.

before | after
----------|----------
image | image
raster: 0.28ms; shadows: 0.48ms | raster: 0.24ms; shadows: 0.42ms

PR #1054, opened 2026-05-29
 demo: Implement hierarchy construction for clusterlod.h

zeux:
As an alternative to flat cluster culling (where for each cluster the
error of cluster vs group is evaluated), we now support BVH based culling
and level selection where the same decisions are evaluated
hierarchically.

`clodBuildHierarchy` builds a forest of trees: each level is represented as
a tree with the requested width that's organized spatially. The tree roots
are located in the first nodes of the output array.

A recursive traversal can descend into the tree while `clodNode::bounds`
is *over* error threshold; this traverses down to leaves, where each
leaf corresponds to a group of clusters; all clusters in the leaves that
have `refined = -1` or `groups[cluster.refined].simplified` at or under
error threshold should be rendered, which creates the correct DAG cut.

~Note that the root is flat for now; there are other ways to construct
the forest but it's likely that an optimal runtime implementation should
special-case the top level traversal anyway, which is made easier by the
flat root as the application could extract errors from nodes 1..level_count
and store them separately to simplify traversal. Nanite uses a slightly
deeper hierarchy with only the coarsest 3 levels at the top, which still
produces a BVH4 but requires a little more traversal to get to more
detailed levels.~

> The forest structure and the construction algorithm follows the
> approach used in NVIDIA's [vk_lod_clusters](https://github.com/nvpro-samples/vk_lod_clusters).

### DIFF
--- a/demo/clusterlod.h
+++ b/demo/clusterlod.h
@@ -137,6 +137,21 @@ struct clodGroup
 // returned value gets saved for clusters emitted from this group (clodCluster::refined)
 typedef int (*clodOutput)(void* output_context, clodGroup group, const clodCluster* clusters, size_t cluster_count);
 
+// when using BVH to determine cluster visibility, traverse the tree into nodes while bounds is over error threshold; for leaves,
+// group clusters should be rendered if cluster.refined is -1 *or* clodGroup::simplified for groups[cluster.refined].simplified is at or under error threshold
+struct clodNode
+{
+	// BVH node bounds, reflecting worst case error for all groups in the subtree
+	clodBounds bounds;
+
+	// leaf node: index of the group this node represents (or -1 for internal nodes)
+	int group;
+
+	// internal node: children are at nodes[child_offset..child_offset+child_count)
+	unsigned int child_offset;
+	unsigned int child_count;
+};
+
 #ifdef __cplusplus
 extern "C"
 {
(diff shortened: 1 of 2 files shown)


zeux:
> Note that the root is flat for now; there are other ways to construct the forest...

I'm not super happy with that detail, because it complicates repacking into a fixed-width BVH, and at the same
time is not general enough as it inserts a faux root, so engines that choose a more optimal 1D scan at the root
level would need to remove the entire root structure and remap offsets...

Still not super sure if maybe the code should build the forest more explicitly without a merged root. Ultimately the specific data layout is annoyingly different based on how the engine prefers to structure it (`vk_lod_clusters` uses a very different layout from Nanite for example, and Nanite implementation has comments suggesting they are considering different root level culling altogether). It would be nice if `clusterlod.h` was agnostic of this detail; I'll leave this open for a bit as I think about possible different shapes for this API that could be more generally useful.

zeux:
After mulling it over I've updated the change to only create a pure forest, without a single root node that has all trees as children. This seems to be a little more generally useful. If we support creating a single tree in the future, it could be a separate function or an extra argument here, but in that case it would probably be best to use a Nanite-style tree (root level -> width-1 coarsest levels as direct children, with the remaining child pointing to a full tree built around the finer levels), as it minimizes the traversal depth for far away objects.

pixeljetstream:
Sounds good. Seconding that forest gives most flexibility

PR #1059, opened 2026-06-19
vertexcodec: Optimize SSSE3 decoding

zeux:
This change significantly reworks SSSE3 decoding implementation to increase performance.

The core idea behind this rewrite is the same as the recently-rewritten AVX-512 implementation: on real workloads, our decoding was branch heavy and this branches predicted poorly; so instead of doing a switch dispatch per byte group, we use a unified SIMD decoder that can simultaneously handle any group type in a data-dependent way.

The latency bypass is taken as is from AVX-512 implementation. It requires a fast popcount, so as part of this change we now require SSSE3+POPCNT for the decoding path; CPUs without POPCNT will fall back to scalar decoding. In practice, all AMD CPUs with SSSE3 also have POPCNT, and this only excludes very early Intel CPUs (Core 2/2006 and Penryn/2007) on Intel side; Nehalem (2008) and later CPUs have POPCNT. We check the CPUID bit at runtime for the fallback or rely on compiler having defined the macros for unconditional inclusion.

The most notable change is the emulation of multishift (which was itself used to emulate multigetbits). We don't have it but we can synthesize something similar by using a mode-dependent shuffle/multiply-high: `pmulhuw` allows us to effectively emulate per-lane right shift for 8 lanes, as multiplying by a power of two and then taking the high 16 bits of the product is equivalent to a right shift. We need to do this twice and combine the results, and some variations are possible here; for now we settle on `pmulhuw`x2 + `psllw` + `por`, but this is also doable with `pmulhuw` + `pmullw` + `pand` + `por`, or with two shuffles `pshufb`x2` + `pmulhuw`x2 + `packuswb` - but we make do with one shuffle as it makes things a little faster on early Intel chips.

Byte expand is mostly taken from our earlier SSSE3 code, and would work as is, but we use a different trick to make this a little faster (which buys ~2%): instead of computing the shuffle adjustment for the second half via a table lookup, `psadw` is used to compute `-popcount(mask0)` in the low byte, so a subtract performs the add we need.

With all of this, we end up with mechanically the same change as AVX-512; however, we use more SIMD resources and so by itself, it's not as fast and can end up losing vs branchy implementation on long runs of 0s or 8s which were trivial to decode. So this change also extends the previous very-fast-path that was limited to zero runs on v0, to handle any four consecutive 0-bit or 8-bit groups regardless of codec version. With this, this change seems to never be slower than the original SSSE3 implementation, on either codec version or on any data set. Of course it should be possible to construct an artificial case where the old implementation never hit branch mispredictions and decoded more quickly (e.g. a long run of 1-bit data), but in practice this is a massive win (data in comments below).

Notably this makes AVX-512 a bit faster (due to new fast path) but also shortens the SSSE-AVX512 gap significantly, which now sits at around 10% on Zen 4 - making runtime dispatch for AVX-512 much less of a priority as the default implementation is now quite fast indeed too.

Thanks to Fabian Giesen for productive discussions and suggestions!

### DIFF
--- a/demo/tests.cpp
+++ b/demo/tests.cpp
@@ -623,6 +623,31 @@ static void decodeVertexBitGroupSentinels()
 	assert(memcmp(decoded, data, sizeof(data)) == 0);
 }
 
+static void decodeVertexBitGroupSentinelCount()
+{
+	const unsigned char expected[13 * 4] = {
+	    0xff, 0, 0, 0, 0xfe, 0, 0, 0, 0xfd, 0, 0, 0, 0xfd, 0, 0, 0,
+	    0xfd, 0, 0, 0, 0xfc, 0, 0, 0, 0xfb, 0, 0, 0, 0xfb, 0, 0, 0,
+	    0xfa, 0, 0, 0, 0xfa, 0, 0, 0, 0xf9, 0, 0, 0, 0xf9, 0, 0, 0,
+	    0xf8, 0, 0, 0 //
+	};
+
+	// encodes several 2-bit sentinels including lane 12; lane 3 is clear so bit 30 does not alias bit 0 during counting
+	const unsigned char input[] = {
+	    0xa1, 0xa9,
+	    0x01, 0xfc, 0x3c, 0xcc, 0xcc,
+	    0x01, 0x01, 0x01, 0x01, 0x01, 0x01, 0x01, 0x01, 0x01,
+	    0x00, 0x00, 0x00, 0x00,
+	    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+	    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+	    0x00, 0x00, 0x00, 0x00 //
+	};
+
+	unsigned char decoded[sizeof(expected)];
+	assert(meshopt_decodeVertexBuffer(decoded, 13, 4, input, sizeof(input)) == 0);
+	assert(memcmp(decoded, expected, sizeof(expected)) == 0);
+}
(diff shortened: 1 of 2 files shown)


zeux:
Here are measurements that include the full AVX-512 speedup (including the previous change) vs 1.0 baseline, as well as the SSSE3 speedup which is entirely due to this change:

Screenshot from 2026-06-19 08-10-01

and just the speedup:

Screenshot from 2026-06-19 08-10-50

What's instructive is that before this change, we had excellent performance on synthetic data sets, and glTF-like data would often encode quite well too because it's more intrinsically redundant due to filters deliberately shifting statistics of individual byte streams, and as such would have lower rate of branch mispredictions. But on game engine data that was already packed well, the codec would have to work hard to find savings and that resulted in switching the modes often, causing high rate of branch mispredictions.

The new version isn't fully branchless, but the branches help save time as they skip the entire 4-group branchless sequence when a consistent run is found, and it's more likely that a branch like this predicts well too.

As one example not captured in this data, `niagara` scene cache on Bistro used to decode vertex data at ~3.4 GB/s - quite a bit lower than the old decoder ran on synthetic data! With this change, it decodes at ~4.9 GB/s which is a much stronger result.

Further polish might be possible in the future although it is getting very difficult due to a variety of limits. AVX-512 now has a very consistent ~10% edge over the SSSE3 decoder whereas previously results were wildly different - even before the recent AVX-512 optimization, the decoder would be branchless only over 1/2/4 groups, so it would have a significant edge where these groups diverge but lose that edge if switching frequently between 4/8 groups for example. Now it's a consistent and relatively modest advantage instead, as both are fully branchless minus the very-fast-path.

PR #1070, opened 2026-07-29
demo: Rework remeshing demo with GPU repainting

zeux:
This is pretty much a complete rewrite of the repainting portion of the remeshing demo :) I got a little carried away...

The original demo was quite slow when generating high resolution textures. Most of the cost was in JS raycasts; I tried a few different options for how to accelerate these (rewrite casting function to iterative traversal; tweak three-mesh-bvh construction; replace three-mesh-bvh with tinybvh compiled to JS; use WebWorkers). Ultimately it was all pretty cumbersome, and the only option that got reasonable tracing times involved tinybvh compiled to JS and used from WebWorkers, which required someone to maintain tinybvh.js and that wasn't going to be me. Plus even once tracing cost was removed, material evaluation was not free and computing normal maps added even more time.

So I decided to bite the bullet and move repainting *entirely* to WebGL. Thankfully three-mesh-bvh was a huge help here, as it already provided full support and shader code for tracing the BVH, including closest-point sampling, and also had helpful utilities for exposing vertex data as textures.

To be able to do material sampling, not just ray casts, materials need to get bound into the shader in a "bindless" manner. Neither WebGPU nor WebGL support bindless, so all textures referenced by all materials need to get rescaled to the same size and copied into a texture array. We limit the size of the resulting array to 1024x1024 layers (and 1 GB total); this limits the quality a little bit for some assets but makes this practical. If WebGPU gains true bindless support in the future this could get ported to WebGPU instead, but for now WebGL is enough.

With all that, ray casting and sampling becomes very fast and is practical to run at high resolutions like 2048. Which then made it fairly easy to integrate normal map and material (metallic-roughness) map baking too. For normal map baking, we need source and destination tangents, and we use MeshoptTangents module to generate these now.

Vertex color repaint works using the same approach, with an intermediate texture that is copied to vertex color stream on the CPU.

The results are quite faithful with the right settings; e.g. this is a remesh of a backpack mesh with normal/material map baking, voxel resolution 128, thickening, simplification enabled and regularization disabled:

before | after | wireframe
----|----|----
image | image | image

The demo now also has an 'exportFile' button which produces a GLB file with all textures bundled, courtesy of three.js GLTFExporter.

The JS code used averaging of pixels along triangle edges; to keep this PR a little smaller we don't do that (it requires a separate intermediate texture with RGBA16F format and a separate resolve shader). There might also be occasional issues with the ray casting code as the thickness needs to be tuned better for some meshes to work well, and I'm not sure if casting is watertight either.

### DIFF
--- a/demo/remesh.html
+++ b/demo/remesh.html
@@ -90,7 +90,7 @@
 			import { MeshBVH } from 'three-mesh-bvh';
 			import { MeshoptDecoder } from '../js/meshopt_decoder.mjs';
 			import { MeshoptSimplifier } from '../js/meshopt_simplifier.js';
-			import * as retexture from './retexture.js';
+			import * as repaint from './repaint.js';
 			import { Initialize as initializeWatlas, Atlas } from 'watlas';
 			import { GUI } from 'lil-gui';
 
@@ -289,7 +289,7 @@
 						var geometry = new THREE.BufferGeometry();
 						geometry.setAttribute('position', new THREE.BufferAttribute(positions, 3));
 
-						// we need original normal/uv/colors for gpu retexture
+						// we need original normal/uv/colors for gpu repaint
 						// for mergeGeometries to agree to merge everything, we maintain consistent attribute sets
 						geometry.setAttribute('normal', copyAttribute(object.geometry.attributes.normal, source.count, 3, 0));
 						geometry.setAttribute('uv', copyAttribute(object.geometry.attributes.uv, source.count, 2, 0));
@@ -674,14 +674,16 @@
 					remeshMaterial.vertexColors = false;
 
 					if (settings.paint.mode == 'vertex') {
-						transferColors(merged, geometry);
+						if (settings.paint.gpu) repaint.transferColors(renderer, geometry, merged.bvh, merged.thickness, merged.materials);
+						else transferColors(merged, geometry);
+
 						remeshMaterial.vertexColors = true;
 						remeshMaterial.color.set(0xffffff);
 					} else if (settings.paint.mode == 'texture') {
 						geometry = unwrapMesh(geometry, settings.paint.resolution, settings.paint.gutter);
 						remeshMaterial.color.set(0xffffff);
 						remeshMaterial.map = settings.paint.gpu
-							? retexture.transferTexture(
+							? repaint.transferTexture(
 									renderer,
 									geometry,
 									merged.bvh,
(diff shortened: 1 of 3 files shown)


zeux:
Profile for the backpack remesh above (with cached paint data; the first remesh pays ~170ms extra to build and upload the cached resources); times are in msec:

image


Generally speaking the bottleneck tends to be in UV unwrapping now; I tried to look a little bit into whether unwrapping can be faster by tweaking existing settings but didn't get very far. It's reasonable on smaller mesh resolutions but pretty much requires simplification to be in the pipeline - unwrapping non-simplified remeshed geometry is quite slow in comparison. The resulting UV charts could be better although unwrapping is a super difficult problem, and also is made worse because the remeshed geometry here occasionally has "fins" (two-sided triangles that connect two other pieces of geometry together) which is an artifact of remeshing I haven't figured out how to cleanly fix yet. Definitely some future work remains before the results are devoid of artifacts.

image

zeux:
Ok, hopefully this is the final piece. The original version had an issue with chart growth that I was hoping would not be visible in practice but it was: the code that grew triangles used an unstable edge offset, that guaranteed a 1px distance between offset edge and original edge - but for very thin triangles, this resulted in pretty much arbitrary elongation which then rasterized excess pixels all over other charts, resulting in occasional line of pixel artifacts. This is in addition to other issues with a large offset - we do not currently differentiate between internal and boundary edges of the chart so we technically expand all triangles inward too, which is suboptimal as well.

This is possible to fix geometrically with an extra layer of triangles but a bit cumbersome; ideally we'd use conservative rasterization but naturally neither WebGL nor WebGPU expose it. So for now I opted for a much less aggressive expansion of the original triangles, which by itself doesn't fully cover the bilinear footprint, followed by an infill pass - we generate mipmaps of the rasterized texture and then replace every pixel with alpha=0 with a different pixel that we find by checking the increasing mip levels with linear filtering. The original data here is trivially premultiplied (alpha=1 or 0), so the generated mips are too, and once we find a filtered pixel with alpha > 0 we un-premultiply it. This results in a nice and cheap neighbor infill while staying within WebGL capabilities and not using any complex hierarchical schemes; as a bonus the textures should compress better now, and are opaque.

This does defeat an ability to preserve alpha cutout surfaces but I've already spent 4 days on this rewrite and I am very done :sweat: maybe I'll add proper transparency support in the future. For now this seems to work at the cost of even more code and generates nicely filled textures like this:

image

PR #1072, opened 2026-08-06
 tangentspace: Implement normal generation with optional smoothing

zeux:
This change implements a new experimental function, `meshopt_generateNormals`, which generates normals with a crease angle cutoff followed by optional smoothing.

The crease angle is used to classify edges into soft/hard; we then use the same union-find method as we use in tangent generation to merge triangle corners into groups, and compute a weighted average of face normals. This is conceptually similar to how Blender computes normals, but Blender makes all non-manifold edges hard; we need them to remain soft because among other things they may be produced by remeshing, where a fin triangle connects to the outer surface and outer surface should stay smooth. We currently use the same weighting we use for tangents (angle * edge length product). This is unconventional but it works reasonably well, although we might switch to something like angle * area in the future if needed to stabilize thin triangles better - tangent weighting was motivated by compatibility with MikkTSpace whereas here the choice is arbitrary. A popular angle weighting is strictly worse on irregular geometry in terms of appearance.

Smoothing is using an iterative Laplacian filter; every iteration smooths one extra edge, so on more densely tessellated models smoothing factor may need to be increased to have the overall appearance match a lower-polygon version. Of course smoothing is quite optional, and setting `smoothing` to 0 disables it. Smoothing never smooths across creases, and uses normal alignment to adjust the filter to avoid oversmoothing. I also tried to incorporate edge length into this, which would make the filter more closely match bilateral normal denoising literature, but it didn't seem to significantly affect the results, and some concerns around robustness and mesh scale invariance made me remove this for now. Smoothing behavior might change in the future, just as the triangle weighting might.

The output is similar to `meshopt_generateTangents`: a per-corner normal needs to be either copied to the unindexed mesh, or propagated into the indexed mesh while splitting vertices.

Notably, in addition to cleaner appearance, we end up generating significantly fewer vertices than three.js's `toCreasedNormals` on typical remeshed inputs, as shown in the comparison image. This is due to the fact that three.js uses a pairwise corner crease test: every corner will average all incident triangles with angle below crease threshold. This means that you get multiple different combinations per vertex which produces more vertex copies and leads to more visible seams; this algorithm outputs slightly smoother and more consistent results. Although ultimately all of these are heuristics and either algorithm has cases where it misbehaves where the other one doesn't.

normals-comparison-final


SashaRX:
Hello, @zeux, some asset pipelines optionally treat UV island boundaries as hard normal seams, even when the geometric crease angle would otherwise keep the edge smooth. This keeps normal and tangent discontinuities aligned and can help avoid tangent-space normal-map baking or interpolation artifacts.

Are there any plans for an optional mode that treats UV discontinuities as hard edges—perhaps by accepting UVs or an explicit seam mask—or is this expected to remain a separate preprocessing step before meshopt_generateNormals?

zeux:
That seems a little backwards to me; what is the example of a pipeline that works with a mesh that has UV islands and doesn't have normals that would require this?

The case that I have ran into so far is the reverse: when there is *no* UV discontinuity but there *is* a normal discontinuity, you get a discontinuous tangent space but the UV is continuous, which is a problem because the texels shared by the edge between two triangles belong to two conflicting tangent spaces.

SashaRX:
You're right—the strict requirement is hard normal edge → UV seam, not the reverse. I was thinking of an optional hard-surface baking workflow where existing UV island boundaries can also constrain normal smoothing. This reduces the strong gradients the normal map must encode across large angles and makes the result more robust to low resolution, BC compression, mipmapping, and LODs. For actual hard normal edges, the UV split is also required to give both smoothing groups independent padding; otherwise filtering and mip generation may sample texels encoded for the tangent frame on the opposite side. So this would not be required for correctness, but an optional UV-seam or explicit edge-mask constraint could be useful.

image
image

zeux:
The distortion in your test is very significant; I would say in this case I'd expect different normals to begin with - it looks like maybe the generated normals here don't take triangle area into account so the normals on the bevels are pointing much more to the side than they should be?

I guess there's a bit of a question as to whether UVs are expected to be hand-authored or not. For automatically unwrapped UVs the problem is that it's quite typical to see UV seams across a smooth surface. This happens often when using xatlas, but in general automatic UV unwrapping is difficult to get seamless. If we assume UVs *are* automatically generated then I worry that making these edges hard will actually introduce more visible seams. If the UVs are human authored then I am not sure if `generateNormals` is particularly useful anyway :) But I'll think about this more; a quick test says the baked normal map might hide this anyway.

smooth seam normals | hard seam normals | with rebaked normal map
----|----|----
image | image | image

SashaRX:
You're right — I phrased it backwards. I meant that every hard normal edge should have a corresponding UV seam. Otherwise the two sides use different tangent frames while sharing the same texels, so they cannot receive independent padding, and filtering or mip generation can bleed normal data between smoothing groups. Would it be useful for meshopt_generateNormals to expose the generated hard-edge mask, so an unwrap or baking pipeline can create the required UV seams, or is that intentionally outside the scope of this API?

zeux:
I don't really want to add an output mask to this API but note that this is already something that is easy to unambiguously detect in the result; xatlas supports setting `normalSeamWeight` to 1000 which is a "soft" guarantee that normal seams result in UV seams: https://github.com/jpcy/xatlas/blob/master/source/xatlas/xatlas.h#L184

(I say "soft" because I think there might be cases based on my testing where this doesn't always happen, but I think it's due to further processing in xatlas that doesn't always look at normals)

Perhaps a more interesting question is whether a UV chart guided normal generation is simply better overall for normal map baking workflows: I have been assuming that the flow is generate normals => unwrap UVs, but you can also unwrap UVs => generate normals, in which case normal generation can be done without crease detection, simply following the UV charts. But that then is easy to implement with a simple accumulation loop so at this point maybe this is just not the job of `meshopt_generateNormals`.

PR #1083, opened 2026-08-26
remesher: Switch to "pure" meshing with cell deciders

zeux:
This change adjusts the remeshing core algorithm, with some API and behavior consequences. Most significantly, it makes thickening a no-op (`meshopt_RemeshThicken` does nothing), and adds `meshopt_SimplifyPreserveFolds` to ensure simplification after remeshing does not erode thin sheet edges. On top of this the raw remeshed output is often cleaner, removing something like ~1% of triangles for fins/bridges, improving appearance of ridges and reducing self-intersections. Symmetrical models are now remeshed more cleanly. Future improvements are possible and planned within the new framework.

The previous version of mesher used a dual grid (this part is left in tact) and a "collapsed" version of Marching Cubes. Which was discovered independently but I later learned has some parallels to prior work. But crucially, the underlying tables were still raw MC and individual vertices could be "thickened" to result in the MC topology, to make thin sheets thicker and preserve 1D geometry better. However, thickening by itself had many problems (diagonal sheets would often thicken poorly, diagonal wires would not be preserved, many cells would be thickened excessively).

> For posterity, "Dual Marching Cubes: Primal Contouring of Dual Grids" by Schaefer and Warren arrives at a similar meshing approach and discusses sliver removal by forcing `w=0`. The way they build the quadrics is very different because they start from an implicit function, and they are restricted to MC-like topology too because `w=0` is not applied universally. Their handling of thin sheets is also quite different as they don't use occupancy.

While thickening could perhaps be improved a bit with more work, the fact that we had to support original MC topology with arbitrary collapses was blocking many ideas I had for how to improve the remeshed geometry; and some fixes for e.g. fin removal would not work correctly around thickened geometry. So thickening was basically an integral part of the algorithm that in the current form did not make sense. The main problem that it *did* solve was that simplification would erode thin sheets as they topologically look like a sphere, so edge quadrics were absent.

So I decided to bite the bullet and completely capitalize on the "pure" mode that was present in the older code implicitly. We no longer use MC tables, and fully embrace the dual grid meshing - triangles connect voxels, not edges between voxels. This has lots of advantages for future improvements; this change takes a few of them but hopefully it's going to make some future planned improvements easier too.

First, the tables are now manually created and audited; they are much simpler structurally, and with rotational symmetry just 24 base cases are enough (technically, it could be 23 cases, but the axis aligned quad is present in two forms to maintain the same quad diagonal on both sides of thin sheets). Many of them have less ambiguity, and the rules for how the tables should be constructed to guarantee the result is topologically sound (no open edges, every edge has a pair) are much less strict vs full MC, which allows the following tweaks.

0x1d | 0x1d alt | 0x69 | 0x69 alt
------|------|-----|----
image | image | image | image

Second, each cell can now have an alternate configuration which maintains edge pairing. This is inspired by "internal deciders" from the MC literature like MC33, but the mechanics are completely different. We do not use "face deciders"; this might be a bit interesting in the future, but remains to be explored; in our case face deciders might be much simpler to implement too because of the reduced case complexity, for example the "horror" case 0x69 where every face is ambiguous which has a lot of combinations in MC is completely free of face ambiguity in our case and can be decided in isolation.

The alternate configuration is selected by evaluating a cell decider; this change implements two: octant decider (using sub-voxel occupancy bits from the existing point splats) and quadric decider (using quad diagonal error to select one of two triangulations). These might see future improvements or variants; octant decider in particular has some limitations due to the need to preserve thin sheet connectivity and I'm sure it could be improved.

> Notably, with this, `meshopt_remesh(NULL)` returns a slight overestimate, and "real" remesh will often return fewer triangles. This was always bound to happen if any forms of triangle pruning were introduced; this only matters for using the native API as the JS wrapper handles everything correctly automatically.

All of this results in cleaner geometry: some fins/bridges that connect adjacent areas of the mesh are now removed, and some ridges now follow the original geometry better:

before | after
-------|------
image | image

To resolve the issue with thin sheet erosion, simplifier now has a "preserve folds" mode which recognizes fold edges using a dihedral angle test, and uses edge quadrics to make sure the quadric error reflects the deformation. This by itself is harmless on pretty much any geometry; but, the extra analysis takes some setup time and makes positional-only simplification ~10% slower give or take. Because it's needed rarely, it's a separate option - it should be a very safe default for remeshing workflows, and can be useful in other cases - for example, this gate from NV Zorah demo has geometry that is dual-sided which erodes under simplification, creating a visible gap (and setting "preserve folds" option fixes that):

Screenshot from 2026-08-23 08-18-01

Finally some other more minor tweaks like solve clamping help reduce geometric artifacts from solve in complex case, and mesh grid alignment is now improved which helps produce cleaner remeshed output for models with inherent symmetry.

For now, I am keeping "thicken" option in the API but it does nothing. I want to come back to it and implement it separately, by post-processing the polygonized output to pull the two sides of a thin sheet apart; this is still valuable to do for diagonal sheets, as they do not necessarily have the same triangulation when meshed from both sides, and should work better than legacy edge thickening without creating more geometry. For wires, I am not sure what the end state would be yet, but they didn't *really* work before either :) voxels are hard.


zeux:
Here's an example of a difficult model ([The Forgotten Knight](https://sketchfab.com/3d-models/the-forgotten-knight-d14eb14d83bd4e7ba7cbe443d76a10fd)) that has complex voxel structure *and* a thin sheet, with simplification enabled:

original | thicken| this change
---------|----------------|------------
image | image | image

PR #1086, opened 2026-08-29
demo: Implement assorted remeshing improvements

zeux:
- New `facets` mode assigns vertex colors and normals to triangles. It looks pretty fun and helps when debugging complex geometry.
- Simplification can use `errorAuto` to adjust the error based on current mesh view distance. This assumes 1Kx1K target for simplicity, similarly to `simplify.html`, so could technically over-simplify a bit relative to what is strictly allowed for current view.
- Simplification is now faster because it skips slow `mergeVertices` in favor of fast `MeshoptSimplifier.generatePositionRemap` which can be used as an index buffer in this case.
- xatlas now can be configured to use faster charting by limiting the chart area; I haven't checked but xatlas chart accumulation might be quadratic in triangle count within one chart, and it's definitely too slow for high-poly inputs without this.
- Remeshing now displays the progress label to understand what is taking time.

The diff is unfortunately gnarly because I had to drop indentation level in the `remesh` function as it's now fully async and I was tired of fighting Prettier for line length... ignoring whitespace helps.

### DIFF
--- a/demo/remesh.html
+++ b/demo/remesh.html
@@ -122,6 +122,7 @@
 					enabled: true,
 					regularize: true,
 					solve: true,
+					errorAuto: false,
 					errorThresholdLog10: 2.5,
 					forceTriangleCount: 0,
 				},
@@ -190,6 +191,7 @@
 			guiSimplify.add(settings.simplify, 'enabled');
 			guiSimplify.add(settings.simplify, 'regularize');
 			guiSimplify.add(settings.simplify, 'solve');
+			guiSimplify.add(settings.simplify, 'errorAuto');
 			guiSimplify.add(settings.simplify, 'errorThresholdLog10', 1.5, 3, 0.1);
 			guiSimplify.add(settings.simplify, 'forceTriangleCount');
 
@@ -349,6 +351,7 @@
 
 				var result = mergeGeometries(geometries);
 				result.computeBoundingBox();
+				result.computeBoundingSphere();
 				result.materials = materials;
 				result.thickness = result.boundingBox.getSize(new THREE.Vector3()).length() * 0.01;
 				return result;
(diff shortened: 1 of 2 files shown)


fallenmi (commented):
Review disclosure: this review was prepared with OpenAI Codex assistance for human review before posting.

The new async progress stages leave the state-changing `remesh`, `revert`, and `loadFile` controls active while `remesh()` is suspended, but the function continues to read the mutable globals after each `await`. In particular, after `await showProgress('remesh')`, `load()` can run and set `model` and `merged` to `undefined`; when the suspended operation resumes, line 514 dereferences `merged.index` and rejects with a `TypeError`. Because the GUI wrapper calls `remesh()` without awaiting or catching it, the rejection is unhandled and the progress label remains visible.

I reproduced this with an exact-source deterministic scheduler that extracts and executes the production `remesh()`, `showProgress()`, and `load()` functions. It invokes the real `load()` at the first newly introduced `requestAnimationFrame` yield:

- exact base `d4c713f2e6a701fecb559d73e32241ccd2040f8a`: remesh completes, one mesh is added, no error;
- exact head `8bfc84c3fb55a58bd66dd4042a0b7b1c44c3a687`: `TypeError: Cannot read properties of undefined (reading 'index')`, no mesh added, progress remains visible;
- exact current merge `a72ad662ac375d7964b26876e0ad22d8d74dd56f`: same failure.

Please bind each run to a captured source/generation and abort cleanly after an `await` if `load()`/`revert()`/another remesh invalidated it (or disable those controls for the full operation), and hide the progress label in a `finally` path. The ordinary `make test` functional run and all 57 JS tests pass on base, head, and merge, and all 13 live checks are green, so they do not cover this browser event-loop interleaving.

PR #1093, opened 2026-09-12
demo: Implement optional border dilation for clusterlod

zeux:
When generating level of detail hierarchy for foliage, the edge collapses by themselves do not leave the LOD that looks similar to the original mesh from a distance. This happens due to a loss of coverage, where at some point the only edge collapses left are ones that would remove individual leaves/grass blades.

What is more problematic is that this results in coordinated omission: across the entire mesh, most blades/leaves are similarly sized; at some point during simplification, many of them will be removed at once, and this will happen at a similar error. The loss of just one leaf is not noticeable, but because the error is similar, many clusters will flip to a coarser LOD at once, causing visible pop - and, eventually, complete erosion.

While this could be mitigated for complex geometry by extruding the borders, the problem with leaves/grass is that each individual feature is tiny, and often it gets collapsed entirely. Thus the only way to prevent this is to redistribute the lost area to other leaves. The new option, `simplify_dilate_borders`, now does that by moving the border vertices outwards to compensate for lost surface area.

> This is more or less the same idea as what UE does for Nanite when "Preserve Surface Area" setting is enabled.

This is intended to be enabled *just* for foliage. However, to mitigate the possible artifacts if enabled broadly, we compute area just for triangles adjacent to border edges; even if this is used on regular meshes, a surface deformation will not cause significant extrusion.

The extrusion is modeled approximately using area/perimeter ratio, with additional miter offset applied but only up to a ~45 degree angle, to prevent sharp angles from excessively protruding outwards.

Since this dilation happens after every simplification step, the subsequent levels of detail get switched correctly, and the dilation itself is gradual as it follows simplification progress.

### DIFF
--- a/demo/clusterlod.h
+++ b/demo/clusterlod.h
@@ -61,6 +61,10 @@ struct clodConfig
 	// try to preserve fold lines between opposite-facing triangles during simplification, at a small performance cost
 	bool simplify_preserve_folds;
 
+	// dilate open borders for each cluster to try to mitigate area loss; most useful for foliage, should not be enabled everywhere
+	// note: this will mutate vertex_positions in place; the caller should copy positions for each cluster in the output callback as they will be overwritten by simplified clusters
+	bool simplify_dilate_borders;
+
 	// should clodCluster::bounds be computed based on the geometry of each cluster
 	bool optimize_bounds;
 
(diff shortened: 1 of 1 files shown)


zeux:
When viewed from up close, the level of detail geometry generated with this option looks pretty strange:

image

... however, it looks completely fine from the correct distance, and this pretty much completely fixes foliage transitions for Zorah (up until quite far away at least) and maintains a reasonable triangle budget at any distance:

image

(the images above are generated with `--simplifynormalweight 1.0 --simplifyuvweight 16.0 --loderrormergeprevious 1.0 --loderrormergeadditive 0.0`, with `simplify_dilate_borders` enabled automatically for meshes with two-sided materials - which in this version works fine because it's being very careful with area estimation)

zeux:
One additional comment is that this code for now is just part of `clusterlod.h`. I've done many other experiments for how to best put this in the simplifier core, but:

- Doing this inside the simplifier is not better other than ease of integration, because extra data it maintains like quadrics or classification does not help that much.
- The cleanest route to supporting this in the library would be a flag that *only* works in `meshopt_simplifyWithUpdate` and *only* is effective if the simplification is iterative, as doing one-shot simplification does not allow the logic to gradually redistribute area. Although maybe we could do this during simplification passes, but that has some other problems with quadric validity.
- Even if it *were* done in `simplifyWithUpdate`, it's currently not integrated with `clusterlod.h` and doing so would require complicated additional changes.

So for now this is *just* code in `clusterlod.h`. I'm also not fond of this method, as it's not very rigorous and can backfire if enabled everywhere, but all alternatives I've tried did not really solve the problem of disappearing foliage. So for now I'm hesitant to add this as a meshoptimizer helper like `meshopt_simplifyDilate` although this certainly could, in theory, happen in the future.

arpu:
Hey what test model is this?

pixeljetstream:
@arpu 
it's from here https://github.com/nvpro-samples/vk_lod_clusters#zorah-demo-scene

